PR #128, opened 2022-10-28
Validate object in `obj.toString()` case

littledivy:
Correctly type check expressions like:
```ts
let a: string = notDefined.toString();
```
```
error[TS2304]: NoSuchVar {
    span: Span {
        lo: BytePos(
            18,
        ),
        hi: BytePos(
            28,
        ),
        ctxt: #0,
    },
    name: notDefined#1,
}
 --> a.ts:1:17
  |
1 | let a: string = notDefined.toString();
```

This is one of the false positive I'm hitting to get `tests/conformance/controlFlowNullishCoalesce.ts` to work.


kdy1 (approved):
Thank you! You are so fast!

PR #129, opened 2022-10-28
Check reserved identifiers in function params

littledivy:
Adds TS1359 validation for function parameter identifiers.

```
error[TS1359]: ExpectedIdentifierReservedKeyword {
    span: Span {
        lo: BytePos(
            21,
        ),
        hi: BytePos(
            26,
        ),
        ctxt: #0,
    },
}
 --> a.ts:1:20
  |
1 | async function foo(await): Promise<void> {
  |                    ^^^^^
```

New passing conforming tests:
```
async/es2017/functionDeclarations/asyncFunctionDeclaration5_es2017.ts
```

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/function/mod.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/function/mod.rs
@@ -3,7 +3,8 @@ use std::borrow::Cow;
 use itertools::{EitherOrBoth, Itertools};
 use rnode::{Fold, FoldWith};
 use stc_ts_ast_rnode::{
-    RBindingIdent, RFnDecl, RFnExpr, RFunction, RIdent, RParamOrTsParamProp, RPat, RTsEntityName,
+    RBindingIdent, RFnDecl, RFnExpr, RFunction, RIdent, RParam, RParamOrTsParamProp, RPat,
+    RTsEntityName,
 };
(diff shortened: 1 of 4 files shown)


crates/stc_ts_file_analyzer/src/analyzer/function/mod.rs
```diff
@@ -478,6 +480,25 @@ impl Analyzer<'_, '_> {
     }
 }
 
+// Check for reserved words used as identifiers (yield, await, etc.)
+//
+// TODO(@littledivy): Generalize this to work regardless of 'function' context.
+fn validate_parameter_name(ctx: &mut Analyzer, p: &RParam, is_async: bool, is_generator: bool) {
```
littledivy, line 486: This should be made context-aware so it can be used with in top level and not just function params. Maybe in the parser? I saw @kdy1 that you have a similar `TODO` a few lines above.

Similar to https://github.com/microsoft/TypeScript/blob/3d2b4017eb6b9a2b94bc673291e56ae95e8beddd/src/compiler/binder.ts#L2178

kdy1: I think it's better to make a method of `Analyzer` for consistency.
I may refactor it to split the analyzer, but let's use a method at now

kdy1:
Also, I think this should be handled by the parser.

PR #139, opened 2022-10-29
chore(ts/checker): Enable `classAbstractKeyword/classAbstractAccessor…

littledivy:


crates/stc_ts_type_checker/tests/tsc.rs
```diff
@@ -216,6 +216,9 @@ fn create_test(path: PathBuf) -> Option<Box<dyn FnOnce() + Send + Sync>> {
 
     if let Ok(errors) = load_expected_errors(&path) {
         for err in errors {
+            if err.code == "TS1318" {
```
kdy1, line 219: I think this line is doing a thing identical to the line below

littledivy: The line below ignores the test `return None`. This line prevents the test from being ignored.

kdy1: My bad, thanks!

PR #158, opened 2022-10-31
fix(ts/analyzer): Fix error message for duplicate class names

littledivy:


kdy1 (approved):

littledivy:
The per test file `tsc-stats` are very useful!

kdy1:
Yeah, I also think so.
I tried generating one because I wanted to spot flaky tests out, but I didn't remove it because I can check regression of each file using it 👍 

And thank you so much for the contributions

PR #167, opened 2022-11-01
feat(ts/analyzer): Improve type fact detection

kdy1:
**Description:**

**Related issue:**


kdy1:
Closing as this only causes regression

Issue #179, opened 2022-11-03
Change error code for using `module` as LHS


kdy1:
https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_type_checker/tests/conformance/expressions/valuesAndReferences/assignments.ts#L10-L11

This line should fail with `TS2708`, but it fails with `TS2539`.

<img width="821" alt="image" src="https://user-images.githubusercontent.com/29931815/199719615-418756b9-a068-493f-bdb5-ae8e15eac08a.png">

The problematic code is at

https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs#L401-L417

and we should use `Errorr::convert_err` to adjust the error variant.

hyp3rflow:
@kdy1 Hi! I'm trying to work on this, I found that the error list provided in stc conformance test is different from typescript's one. Is it okay to be different? Thanks!
[assignments.errors.txt in TypeScript repo](https://github.com/microsoft/TypeScript/blob/main/tests/baselines/reference/assignments.errors.txt)
[assignments.errors.json in stc repo](https://github.com/dudykr/stc/blob/main/crates/stc_ts_type_checker/tests/conformance/expressions/valuesAndReferences/assignments.errors.json)

hyp3rflow:
> @kdy1 Hi! I'm trying to work on this, I found that the error list provided in stc conformance test is different from typescript's one. Is it okay to be different? Thanks! [assignments.errors.txt in TypeScript repo](https://github.com/microsoft/TypeScript/blob/main/tests/baselines/reference/assignments.errors.txt?rgh-link-date=2022-11-03T14%3A37%3A21Z) [assignments.errors.json in stc repo](https://github.com/dudykr/stc/blob/main/crates/stc_ts_type_checker/tests/conformance/expressions/valuesAndReferences/assignments.errors.json?rgh-link-date=2022-11-03T14%3A37%3A21Z)

Oh I found that the conformance test case in stc was outdated. I'll patch it together :)

kdy1:
~We should update tests, but for ~

Can you split the PR for updating tests?

hyp3rflow:
@kdy1 okay. first of all, is there any script for updating errors.json file for conformance test? If not, I'll update it manually (there are 7-8 test cases need to be updated)

[...]

PR #187, opened 2022-11-04
Update TypeScript tests and fix outdated test update scripts

hyp3rflow:
**Description:** Update tests: https://github.com/dudykr/stc/issues/179#issuecomment-1302222809

**BREAKING CHANGE:** Some tests..?


hyp3rflow:
Hi @kdy1! I did update tests as you said and it makes enormous amount of diff on test :|
Is it okay to go? If there are any step to go further or ignore something, feel free to tell me :)

kdy1:
I think it's fine, we should update it anyway. Thank you!

hyp3rflow:
@kdy1 it seems that this PR adds `.stats.rust-debug` file for d.ts test case which is ignored by gitignore. Is there any context for ignoring `*.d.ts`? If it exists, I will open new PR for removing that files.

kdy1:
`.d.ts` is ignored for historical reasons, but it's not relevant anymore so I think we should remove it from `.gitignore`

hyp3rflow:
@kdy1 then I will open new PR for it!

kdy1:
Thank you!

PR #191, opened 2022-11-04
fix(ts/analyzer): remove unused code in the class for simplifying Union Type

### LINKED ISSUE #175: Append `undefined` to type list on optional member access
https://github.com/dudykr/stc/blob/e69d14c1b30473f73bc8873400e950d44a047e64/crates/stc_ts_file_analyzer/tests/pass/exprs/optionalChaining/.access-6.ts#L2-L3

The type of `o6<number>()?.["x"];` above should be `number | undefined`, but currently it's `number`.

Relevant code for optional chaining is  https://github.com/dudykr/stc/blob/e69d14c1b30473f73bc8873400e950d44a047e64/crates/stc_ts_file_analyzer/src/analyzer/expr/optional_chaining.rs#L25-L49

nissy-dev:
Fix #175 

**Description:**

I remove unused code in the class for simplifying Union Type and add the test case.

**BREAKING CHANGE:**


kdy1:
Van you run check.sh to update stats?

crates/stc_ts_file_analyzer/src/analyzer/generalize.rs
```diff
@@ -72,7 +72,6 @@ impl Simplifier<'_> {
 impl Fold<Union> for Simplifier<'_> {
     fn fold(&mut self, mut union: Union) -> Union {
         let should_remove_null_and_undefined = union.types.iter().any(|ty| match ty.normalize() {
-            Type::TypeLit(..) => true,
```
nissy-dev, line 75: Currently , I removed only this line. If I can remove `should_remove_null_and_undefined` itself, I will add new commit.

kdy1:
You have to run it again

nissy-dev:
> You have to run it again, also, can you run it after rebasing?

Yes. there is no additional differences.

crates/stc_ts_file_analyzer/src/analyzer/generalize.rs
kdy1: I think we should remove all of them

kdy1: Can you remove all of this?

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/generalize.rs
nissy-dev: I removed!

kdy1 (approved):

kdy1:
Thanks!

PR #192, opened 2022-11-04
fix(ts/analyzer): Fix wrong error code when assigning to non-variable

### LINKED ISSUE #179: Change error code for using `module` as LHS

hyp3rflow:
**Description:** Fix #179 

**BREAKING CHANGE:**

**Related issue (if exists):** #179

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/control_flow.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/control_flow.rs
@@ -875,6 +875,7 @@ impl Analyzer<'_, '_> {
                                     return Err(Error::NotVariable {
                                         span: i.id.span,
                                         left: lhs.span(),
+                                        ty: Some(box ty.normalize().clone()),
                                     });
                                 }
                                 _ => {}
--- a/crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
@@ -412,8 +412,8 @@ impl Analyzer<'_, '_> {
                                     Some(v) if v.kind == VarKind::Fn => Error::CannotAssignToFunction { span },
                                     _ => err,
                                 },
-                                Error::NotVariable { .. } => match analyzer.type_of_raw_var(i, TypeOfMode::LValue) {
-                                    Ok(ty) => match ty.normalize() {
+                                Error::NotVariable { ty, .. } => match ty {
+                                    Some(ty) => match **ty {
                                         Type::Module(..) => Error::CannotAssignToModule { span },
                                         Type::ClassDef(..) => Error::CannotAssignToClass { span },
                                         Type::Enum(..) => Error::CannotAssignToEnum { span },
@@ -2901,7 +2901,11 @@ impl Analyzer<'_, '_> {
         let mut modules = vec![];
         let mut ty = self.type_of_raw_var(i, type_mode)?;
         if type_mode == TypeOfMode::LValue && (ty.is_class_def() || ty.is_enum_type()) {
-            return Err(Error::NotVariable { span, left: span });
+            return Err(Error::NotVariable {
+                span,
+                left: span,
+                ty: Some(box ty.normalize().clone()),
+            });
         }
         ty.assert_valid();
         if let Some(type_args) = type_args {
@@ -2934,7 +2938,13 @@ impl Analyzer<'_, '_> {
             if let Some(types) = self.find_type(self.ctx.module_id, &id)? {
                 for ty in types {
                     match ty.normalize() {
-                        Type::Module(..) => return Err(Error::NotVariable { span, left: span }),
+                        Type::Module(..) => {
+                            return Err(Error::NotVariable {
+                                span,
+                                left: span,
+                                ty: Some(box ty.normalize().clone()),
+                            })
+                        }
                         _ => {}
                     }
                 }
@@ -3089,7 +3099,11 @@ impl Analyzer<'_, '_> {
         match i.sym {
             js_word!("undefined") => {
                 match type_mode {
-                    TypeOfMode::LValue => self.storage.report(Error::NotVariable { span, left: span }),
+                    TypeOfMode::LValue => self.storage.report(Error::NotVariable {
+                        span,
+                        left: span,
+                        ty: None,
+                    }),
                     TypeOfMode::RValue => {}
                 }
                 return Ok(Type::undefined(span, Default::default()));
--- a/crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
@@ -421,7 +421,10 @@ impl Analyzer<'_, '_> {
                                     },
                                     _ => err,
                                 },
-                                _ => err,
+                                _ => {
+                                    skip_right = false;
+                                    err
+                                }
                             }
                         })
                         .report(&mut analyzer.storage);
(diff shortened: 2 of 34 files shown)


hyp3rflow:
I succeeded in passing the `assignments.ts` conformance TC, but it seems that this change occurs additional error on another TCs. I'll check this further.

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -405,14 +405,24 @@ impl Analyzer<'_, '_> {
...
-                                }
-                                _ => {}
+                        .convert_err(|err| {
+                            skip_right = true;
+                            match err.actual() {
+                                Error::CannotAssignToNonVariable { .. } => match analyzer.scope.vars.get(&i.into()) {
+                                    Some(v) if v.kind == VarKind::Fn => Error::CannotAssignToFunction { span },
+                                    _ => err,
+                                },
+                                Error::NotVariable { .. } => match analyzer.type_of_raw_var(i, TypeOfMode::LValue) {
```
littledivy, line 415: This is not good for performance. The extra `type_of_var` call is wasting a lot of cycles here. I think `Error::NotVariable` should be modified to contain `ty` it got.

hyp3rflow: Thanks! I fixed this as you said, but I'm not sure I did it right. Could you review my fix? [`8507c7b` (#192)](https://github.com/dudykr/stc/pull/192/commits/8507c7b3075a853907dceaf800e09fc90a0fddb7)

littledivy: LGTM! cc @kdy1

crates/stc_ts_file_analyzer/src/analyzer/control_flow.rs
```diff
@@ -875,6 +875,7 @@ impl Analyzer<'_, '_> {
                                     return Err(Error::NotVariable {
                                         span: i.id.span,
                                         left: lhs.span(),
+                                        ty: Some(box ty.normalize().clone()),
```
kdy1, line 878: Please change this to `ty.clone()`

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -405,14 +405,27 @@ impl Analyzer<'_, '_> {
...
-                                    skip_right = true;
+                        .convert_err(|err| {
+                            skip_right = true;
+                            match err.actual() {
+                                Error::CannotAssignToNonVariable { .. } => match analyzer.scope.vars.get(&i.into()) {
+                                    Some(v) if v.kind == VarKind::Fn => Error::CannotAssignToFunction { span },
+                                    _ => err,
+                                },
+                                Error::NotVariable { ty, .. } => match ty {
+                                    Some(ty) => match **ty {
```
kdy1, line 416: Please change this to `match ty.normalize() { ... }`

kdy1: In this way we can avoid cloning

kdy1 (changes requested):
Thank you so much!
Other changes look good to me

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -2891,7 +2904,11 @@ impl Analyzer<'_, '_> {
         let mut modules = vec![];
         let mut ty = self.type_of_raw_var(i, type_mode)?;
         if type_mode == TypeOfMode::LValue && (ty.is_class_def() || ty.is_enum_type()) {
-            return Err(Error::NotVariable { span, left: span });
+            return Err(Error::NotVariable {
+                span,
+                left: span,
+                ty: Some(box ty.normalize().clone()),
```
kdy1, line 2910: Oh same here

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -2924,7 +2941,13 @@ impl Analyzer<'_, '_> {
             if let Some(types) = self.find_type(self.ctx.module_id, &id)? {
                 for ty in types {
                     match ty.normalize() {
-                        Type::Module(..) => return Err(Error::NotVariable { span, left: span }),
+                        Type::Module(..) => {
+                            return Err(Error::NotVariable {
+                                span,
+                                left: span,
+                                ty: Some(box ty.normalize().clone()),
```
kdy1, line 2948: Ditto

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/control_flow.rs
hyp3rflow: Well, how do I have to type `Error::NotVariable`?
Currently the `ty` is Option<Box<Type>>, it makes conflict with this changes.

hyp3rflow: fyi. ty is `Cow<Type>` in this file.

kdy1: Ah, I meant removing `normalize()` only

kdy1: Actually we use `Type::Arc` to avoid cloning and `normalize` dereferences `Type::Arc` into actual data type. So we should not call `.normalize().clone()` because it will clone actual data instead of `Arc`

kdy1: Oh.. `Cow` is problematic at the moment, I'll optimize this in future.  Thank you!
Anyway, please change match to call `normalize()` as it will guarantee that it'll work in future

hyp3rflow: Thanks! `normalize()` at the use-side, right? I'll change it. @kdy1

kdy1: Yup

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
hyp3rflow: ty is `Cow<Type>` in here, too!

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
hyp3rflow: ✅ [`43a0b63` (#192)](https://github.com/dudykr/stc/pull/192/commits/43a0b63877b6837b2505f8695264226b0c134fdc)

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
hyp3rflow: ✅ [`43a0b63` (#192)](https://github.com/dudykr/stc/pull/192/commits/43a0b63877b6837b2505f8695264226b0c134fdc)

kdy1 (approved):
Thanks!

PR #193, opened 2022-11-04
fix(ts/analyzer): change error code for missing new for classes 

### LINKED ISSUE #178: Change error code for missing `new` for classes
https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_type_checker/tests/conformance/expressions/functionCalls/forgottenNew.ts#L1-L5

The code above should fail with `TS2348`, but currently it failes with `TS2349`

image

It has the context of `tried to calculate return type`, which is used at

https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs#L1973-L1977

But error is `NoCallablePropertyWithName`, which is `TS2349`, and

image

---

So we should modify

https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs#L821-L825

to use different error variant.

mrkldshv:
**Description:**

Add new error `NoConstructablePropertyWithName` that maps to `TS2348` code, and change the error conversion in for classes creation with `new`.

**BREAKING CHANGE:**

**Related issue (if exists):** Fixes #178


mrkldshv:
I see there're some tests not related to the changes that are failing. Do we need to do anything extra here?

kdy1:
I'll update the test list to ignore flaky tests first

kdy1:
>  I see there're some tests not related to the changes that are failing. Do we need to do anything extra here?

It seems like those failures are regressions caused by this PR

kdy1 (dismissed):

kdy1 (approved):

Issue #185, opened 2022-11-04
Rest pattern should not trigger `TupleIndexErrror`


kdy1:
https://github.com/dudykr/stc/blob/2ab93ad46cd2683539300bde8f30c7bf89ac1e15/crates/stc_ts_type_checker/tests/conformance/es6/destructuring/declarationsAndAssignments.ts#L180

The line above should not error because LHS is a rest pattern.
<img width="732" alt="image" src="https://user-images.githubusercontent.com/29931815/199899870-5138cdb8-832b-4ca0-9f8a-605655413646.png">

But currently stc fails with an error, because of the code below

https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L250-L263

This is quite similar to https://github.com/dudykr/stc/issues/182, though.

AlexMunoz:
Hi @kdy1, I'd love to work on this issue. Do you have any additional advice? Thanks.

kdy1:
@AlexMunoz Basically, the problem is that we are returning errors from

```rust
             .with_context(|| { 
                 format!( 
                     "tried to get the type of {}th element from iterator to declare vars with an array pattern", 
                     idx 
                 ) 
             })? 
```

using `?` token. Instead of returning an error, we should recover from the error, based on the variant.
`Error#actual()` gives you the actually error value without context, so you can match on it.

AlexMunoz:
Hey @kdy1 I think the PR https://github.com/dudykr/stc/pull/202  is also covering this case.

kdy1:
Oh maybe. My bad, sorry

PR #202, opened 2022-11-05
Fix error reporting for array pattern elements with default values

### LINKED ISSUE #182: Don't report an error on array pattern elements with a default value
https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_type_checker/tests/conformance/es6/destructuring/destructuringArrayBindingPatternAndAssignment3.ts#L1

This should not error, but currently, stc emits `TupleIndexError`

image

---

We should recover from the error based on the variant at https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L250-L263

ever0de:
**Description:** 
Fix to not report errors for array pattern elements with default values.
Since the existing tsc was holding `Any`, it was changed to report as `Any`.
image
-> 
image

**BREAKING CHANGE:**
The phenomenon reported as an error is changed to ‘Type’ conversion.

 - fixes #182 
 - Fixes https://github.com/dudykr/stc/issues/185


crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
```diff
@@ -249,16 +249,27 @@ impl Analyzer<'_, '_> {
...
+                                            idx
+                                        )
+                                    });
+
+                                    match result {
+                                        Ok(ty) => Ok(ty.into_owned()),
+                                        Err(err) => match err.actual() {
+                                            Error::TupleIndexError { .. } => {
+                                                if let RPat::Assign(prop) = elem {
+                                                    Ok(Type::any(prop.span, Default::default()))
```
kdy1, line 265: I'm not sure `any` is correct here

kdy1: This is wrong. We should use `default_ty` instead of `any`

kdy1:
```ts
const [a, b = 1] = [1]; // ok
```

`b` should be number (generalized from 1) here, and this method accepts `default_ty` which can be used for it

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
ever0de: Oh sure..! Okay. I'll edit it and post it. thank you 🙇

ever0de: Oh, the first time I tried this, it failed.
When I took a picture of the current [L232 default](https://github.com/dudykr/stc/pull/202/files#diff-7b41dc3d40058bc0dceb3410349aab0fee895bf7ca944ee6a8c7766faea874dfL232), 
In the case of the current case, ‘None’ is displayed.

Maybe I got `default` wrong?

```
default: None
   WARN ((1:0)): Statement validation done. (time = 1.891458ms, kind: "perf", op: "validate (Stmt)"
    at crates/stc_ts_file_analyzer/src/analyzer/stmt/mod.rs:66
    in validate<RStmt>
    in validate<RModuleItem>
    in validate<RModule>

thread 'pass_tests__pass__exprs__array__destructuring_ts' panicked at 'Failed to validate.
error[TS2493]: DebugContext(
    context: tried to get the type of 1th element from iterator to declare vars with an array pattern
    context: tried to access property of a type to calculate element type
    context: tried to access property of an object ([1];, id_ctx = Var)
    Prop=Num(RNumber { span: Span { lo: BytePos(7), hi: BytePos(17), ctxt: #0 }, value: 1.0 })
    context: r-value context
    TupleIndexError {
        span: span {
            lo: BytePos(
                7,
            ),
            hi: BytePos(
                17,
            ),
            ctxt: #0,
        },
        len: 1,
        index: 1,
    },
)
 --> /tests/pass/exprs/array/destructuring.ts:1:7
  |
1 | const[a, b = a] = [1]; // ok
```

ever0de: Oh I think I found it above.
https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L149

It looks like it could be done by making `right` the default_ty, I'll fix it once

kdy1: You should recover from https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L253-L260

In the closure, you can check if `default` has a useful value

ever0de: But there is no `default` in the current situation (`None` printed in the comment above).
Perhaps the default_type is unknown, and the type must be inferred from the value of `Pat::Assign(assign_pat)`, `AssignPat::right`, am I wrong?

Thanks as always for your replies!

kdy1: You should use `default_elem_ty` below. I think you can change the order of statements.
Also, I think `default` should not be `None`. If it's `None`, I think it's a bug.

ever0de: https://github.com/dudykr/stc/pull/202#discussion_r1014600968
In the comment, I did not take the `default_elem_ty` you mentioned, but since `default` is already `None`, `default_elem_ty` will also be `None`.

ever0de: How do you debug in this regard?
Right now, I am printing a variable with `println`, but there are many stdouts that are output, so it is difficult to understand quickly..

kdy1: Oh, my bad. It should be computed from `RPat::Assign` as you said. `default` is for

```js
const [[a, b] = [1, 2]] = []
```

where the handler for `a` or `b` does not have access to `[1, 2]`

kdy1: I typically use `dbg!` macro and `print_backtrace()` to debug.
As I reduce tests to very small cases, those were enough.

ever0de: So now I'm trying to do something like this.
https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L149-L162
It seems that the above is a way to infer the type of `RAssignPat::right`.

By matching this in `Result<Cow<Type>, Error>` made of `get_element_from_iterator` + `with_context`
If it is `TupleIndexError`, it tries to infer the type of `RAssignPat::right` and pass it.

how about it?

kdy1: I think it's correct, but I think we should unconditionally invoke `right.validate_with_args` and use the result only if required, so the RHS of an assignment pattern can be validated regardless of tuple index error.

ever0de: > use the result only if required, so the RHS of an assignment pattern can be validated regardless of tuple index error.

I didn't quite understand this, but I tried to fix it first.

kdy1 (approved):
Thank you!

PR #209, opened 2022-11-05
Fixes #205

### LINKED ISSUE #205: Implement type inference for tuple and `Array` interface
# Problem

https://github.com/dudykr/stc/blob/d2628d0e537ca1df62829b1c2ef99d898aae1ffa/crates/stc_ts_type_checker/tests/conformance/types/literal/numericStringLiteralTypes.ts#L24-L40

This should work, but it fails with an error

image

It's because the type inference for

```
param  = [...T#11#0];
arg = Array#0;
```

is not implemented.

image

https://github.com/dudykr/stc/blob/d2628d0e537ca1df62829b1c2ef99d898aae1ffa/crates/stc_ts_file_analyzer/src/analyzer/generic/mod.rs#L1285-L1290

# Solution

We should add an inference logic at above of the `error!`

I'm marking this as a good-first issue because there's a very similar code.

https://github.com/dudykr/stc/blob/d2628d0e537ca1df62829b1c2ef99d898aae1ffa/crates/stc_ts_file_analyzer/src/analyzer/generic/mod.rs#L761-L797

Actually, Array-Tuple is implemented, but problem is that Tuple-Array (interface) is not implemented.

AcrylicShrimp:
**Description:**
This issue handles #205 by implementing assign from Array to single-element Tuple of rest.

Fixes #205

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
@@ -5,7 +5,7 @@ use stc_ts_errors::{debug::dump_type_as_string, DebugExt, Error};
 use stc_ts_file_analyzer_macros::context;
 use stc_ts_types::{
     Array, Conditional, EnumVariant, Instance, Interface, Intersection, Intrinsic, IntrinsicKind, Key, KeywordType, KeywordTypeMetadata,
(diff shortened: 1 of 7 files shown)


kdy1:
Can you sign the CLA?

kdy1 (dismissed):
Thanks!

AcrylicShrimp:
> Can you sign the CLA?

Of course!

kdy1 (approved):

PR #221, opened 2022-11-06
fix(ts/checker): Default value of an object pattern property should not change type if it's assignable

### LINKED ISSUE #189: Default value of an object pattern property should not change type if it's assignable
https://github.com/dudykr/stc/blob/d121f20feb4f7a66e96a1b093dfadc507baf8610/crates/stc_ts_type_checker/tests/conformance/types/nonPrimitive/nonPrimitiveAndEmptyObject.ts#L6-L17

This should success.

image

`access_property` of `((BarProps#1 & object) | {})` is expected to fail, because `{}` does not have the property.
But the type of `fooProps` should be `(BarProps#1 & object)` (without union)

I think the logic at https://github.com/dudykr/stc/blob/d121f20feb4f7a66e96a1b093dfadc507baf8610/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L483-L500 is problematic, and I think we should adjust it to ignore default value if the type of the default value is assignable to the original type.

kharvd:
**Description:**

Fix #189

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
@@ -122,7 +122,7 @@ impl Analyzer<'_, '_> {
 
                 let mut ty = match (default, ty) {
                     (Some(default), Some(ty)) => {
-                        if self.extends(span, &default, &ty, Default::default()).is_some() {
+                        if let Some(true) = self.extends(span, &default, &ty, Default::default()) {
                             Some(ty)
                         } else {
                             opt_union(span, Some(ty), Some(default))
--- a/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
@@ -120,7 +120,17 @@ impl Analyzer<'_, '_> {
                     }
                 }
 
-                let mut ty = opt_union(span, ty, default);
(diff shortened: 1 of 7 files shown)


crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
```diff
@@ -120,7 +120,17 @@ impl Analyzer<'_, '_> {
                     }
                 }
 
-                let mut ty = opt_union(span, ty, default);
+                let mut ty = match (default, ty) {
+                    (Some(default), Some(ty)) => {
+                        if self.extends(span, &default, &ty, Default::default()).is_some() {
```
kdy1, line 125: I think `extends().is_some()` is wrong because `extends()` returns `None` only if it cannot be determined. i.e., type parameter

kdy1: I think it should be `if let Some(true) = self.extends(...) {} else {}`

kdy1 (changes requested):
Also, can you sign the CLA?

crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
kharvd: Oops, gotcha, fixed!

kdy1 (approved):
Thanks!

Issue #211, opened 2022-11-05
Overload handling should handle constructor


kdy1:
# Problem

https://github.com/dudykr/stc/blob/269f6c07676f728e667ca5fb4c80d7f17f66a98d/crates/stc_ts_type_checker/tests/conformance/types/objectTypeLiteral/constructSignatures/constructSignaturesWithOverloads2.ts#L15-L24

should pass, but stc fails to verify it.

Currently we only check overloads for methods, so overload selection for `new` fails

# Solution

We should handle constructor from `extract` https://github.com/dudykr/stc/blob/805667830752f229f1674d034e1e9b0eb3fcb076/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs#L1215-L1647

yamgent:
Hi, I would like to take this!

kdy1:
Thanks!

kdy1:
Updated the issue to reflect the real solution

kdy1:
Oh it's a union. `namespace` is rendered as `any` in type dump

yamgent:
Ok, so I tried the new real solution by doing something like this:

```rs
diff --git a/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs b/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
index d655cf6e..523e5722 100644
--- a/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
@@ -1515,20 +1515,19 @@ impl Analyzer<'_, '_> {
                 )
             }
 
-            // Type::Constructor(ty::Constructor {
-            //     ref params,
-            //     ref type_params,
-            //     ref ret_ty,
-            //     ..
-            // }) if kind == ExtractKind::New => self.try_instantiate_simple(
-            //     span,
-            //     ty.span(),
-            //     &ret_ty,
-            //     params,
-            //     type_params.as_ref(),
-            //     args,
-            //     type_args,
-            // ),
+            Type::Constructor(c) if kind == ExtractKind::New => self.get_return_type(
+                span,
+                kind,
+                expr,
+                c.type_params.as_ref().map(|v| &*v.params),
+                &c.params,
+                *c.type_ann.clone(),
+                type_args,
+                args,
+                arg_types,
+                spread_arg_types,
+                type_ann,
+            ),
             Type::Union(..) => {
[...]

PR #222, opened 2022-11-06
fix(ts/analyzer): Overload handling should handle constructor

yamgent:
**Description:**

I am not sure whether my fix is correct, so I am going to describe what I am doing here.

This MR is fixing #211. The original issue's proposed solution was to handle `ClassMember::Constructor` in `call_property_of_class()`. However, when I tried it out, it seems to never call the method.

So I retrace the error and found this, which seems to indicate that it actually does handle the constructor method correctly, and the root cause is that it didn't handle the generic type param `T` properly:

stc-211-error-highlight

Hence I set `allow_assignment_to_param: true` to allow assigning `T = 1`.

#211


kdy1:
Oh yeah it's my bad. The handling logic should be added to `extract` https://github.com/dudykr/stc/blob/805667830752f229f1674d034e1e9b0eb3fcb076/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs#L1215-L1647

crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
```diff
@@ -2920,6 +2920,7 @@ impl Analyzer<'_, '_> {
                                 span: arg.span(),
                                 allow_unknown_rhs,
                                 use_missing_fields_for_class: true,
+                                allow_assignment_to_param: true,
```
kdy1, line 2923: This is wrong

crates/stc_ts_type_checker/tests/conformance/types/typeRelationships/typeInference/contextualSignatureInstantiation.ts.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 0,
-    matched_error: 6,
+    required_error: 3,
```
kdy1, line 2: This is a regression, and to fix it we should not allow assignment to parameters

yamgent:
@kdy1 I think it now excludes the `module C2`. Lemme know if I am still doing something wrong...

crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
```diff
@@ -430,7 +430,30 @@ impl Analyzer<'_, '_> {
...
+
+                        match types.len() {
+                            0 => Type::never(
+                                u.span,
+                                KeywordTypeMetadata {
+                                    common: u.metadata.common,
+                                    ..Default::default()
+                                },
+                            ),
+                            1 => types[0].clone(),
```
kdy1, line 451: Please use `types.into_iter().next().unwrap()` instead, to avoid clone

crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
```diff
@@ -430,7 +430,30 @@ impl Analyzer<'_, '_> {
...
                 }
-                callee_ty
+
+                match callee_ty.normalize() {
+                    Type::Union(u) => {
+                        let types = u
+                            .types
+                            .iter()
+                            .cloned()
+                            .filter(|callee| !matches!(callee.normalize(), Type::Module(..)))
```
kdy1, line 440: Also, please check `Type::Namespace` too.

`module Foo` is `Type::Namespace`, and `namespace Foo` is `Type::Module`

kdy1 (dismissed):
Thank you!

kdy1 (approved):

PR #229, opened 2022-11-06
fix(ts/analyzer): report at max one argument type mismatch error for generic calls

### LINKED ISSUE #223: Report at max one argument type mismatch error for generic calls
# Problem

https://github.com/dudykr/stc/blob/d8bb0e132d8ffdfa898c71165f7777d7bfb24eb3/crates/stc_ts_type_checker/tests/conformance/types/stringLiteral/typeArgumentsWithStringLiteralTypes01.ts#L87-L91

image

Even when generics arguments are specified by the user, `tsc` reports only one argument type mismatch error if it's a generic call.

# Solution

We should accept an argument from `validate_arg_types` to check if it's a generic call.

https://github.com/dudykr/stc/blob/d8bb0e132d8ffdfa898c71165f7777d7bfb24eb3/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs#L2691

After than, if it's a generic call, we should abort on first argument type error

littledivy:
Fixes #223


kdy1 (approved):
Thanks!

Issue #186, opened 2022-11-04
Rest patterns should get the correct type


kdy1:
https://github.com/dudykr/stc/blob/2ab93ad46cd2683539300bde8f30c7bf89ac1e15/crates/stc_ts_type_checker/tests/conformance/es6/destructuring/declarationsAndAssignments.ts#L178

This line should not error, but stc fails to validate it because stc thinks the type of `a1` is `string[]` instead of `[string, boolean]`.

I think the problematic code is https://github.com/dudykr/stc/blob/2ab93ad46cd2683539300bde8f30c7bf89ac1e15/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L292-L326 , which means `get_rest_elements` is returning a wrong type for the input above.

TrickyPi:
Hi @kdy1, i'd like to fix it 😊 .

kdy1:
Thank you!

TrickyPi:
Hi @kdy1

I created a minimal test for debugging
<img width="1189" alt="test" src="https://user-images.githubusercontent.com/33021497/200171433-8baf28f8-50c0-4c02-b459-d181a01a8423.png">
[the file link](https://github.com/TrickyPi/stc/blob/fix/issue-186/crates/stc_ts_file_analyzer/tests/pass/vars/destructuring/array/assignment/21.ts#L1-L5
)

and the error information is
<img width="996" alt="error" src="https://user-images.githubusercontent.com/33021497/200170785-ce5be463-aa20-43de-9323-1e828ae58d05.png">

I found that the error happens here
https://github.com/dudykr/stc/blob/2ab93ad46cd2683539300bde8f30c7bf89ac1e15/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs#L375-L413

Do you have any advice? Seems the bug isn't here
https://github.com/dudykr/stc/blob/2ab93ad46cd2683539300bde8f30c7bf89ac1e15/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L292-L326

BTW, `opts.use_iterator_for_array` is true
https://github.com/dudykr/stc/blob/2ab93ad46cd2683539300bde8f30c7bf89ac1e15/crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs#L210-L372

kdy1:
I guess that `get_rest_elements` is returning the wrong value. 

And there's a TODO for it

[...]

PR #267, opened 2022-11-11
fix(ts/analyzer): Use empty tuple type for rest pattern with tuple initializer

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/dudykr/stc/issues/186.


TrickyPi:
It seems this PR doesn't fix the #186 issue, the type of `a1` is still recognized as `string[]`

Issue #176, opened 2022-11-03
Use `Iterable` instead of `Iterator` for generator functions


kdy1:
https://github.com/dudykr/stc/blob/e69d14c1b30473f73bc8873400e950d44a047e64/crates/stc_ts_file_analyzer/tests/tsc/types/builtin/generator/bad/.1.ts#L1-L2

This should fail because a generator function uses the item from `Iterable` as yield type.
i.e., `g3` is inferred to emit a string as an item, and the inferred generator (`Generator<number, any, undefined>`) is assigned to `BadGenerator`.

It should file with

```
tests/tsc/types/builtin/generator/bad/1.ts:2:24 - error TS2322: Type 'Generator<string, any, undefined>' is not assignable to type 'BadGenerator'.
  The types returned by 'next(...)' are incompatible between these types.
    Type 'IteratorResult<string, any>' is not assignable to type 'IteratorResult<number, any>'.
      Type 'IteratorYieldResult<string>' is not assignable to type 'IteratorResult<number, any>'.
        Type 'IteratorYieldResult<string>' is not assignable to type 'IteratorYieldResult<number>'.
          Type 'string' is not assignable to type 'number'.

```

But currently, stc fails to emit an error because stc uses the `Iterator` interface instead of the `Iterable` interface while inferring the `yield` type.

https://github.com/dudykr/stc/blob/e69d14c1b30473f73bc8873400e950d44a047e64/crates/stc_ts_file_analyzer/src/analyzer/stmt/return_type.rs#L176-L181

kdy1:
Cool, thanks!

kdy1:
@timeTravelCYN Are you working on it? Feel free to tell me if you want me to take it (or simply unassign)

devgony:
Do we have to implement `get_iterable_element_type` from scratch including all functions related with iterable like `get_next_value_type_of_iteratable`?

kdy1:
Not sure, but I think we can reuse many parts of it.
Maybe we can extract some common logic to a function

PR #295, opened 2022-11-15
Handle `intrinsics` for template literal types

### LINKED ISSUE #210: Handle `intrinsics` for template literal types
https://github.com/dudykr/stc/blob/269f6c07676f728e667ca5fb4c80d7f17f66a98d/crates/stc_ts_type_checker/tests/conformance/types/literal/stringMappingOverPatternLiterals.ts#L2-L4

https://github.com/dudykr/stc/blob/269f6c07676f728e667ca5fb4c80d7f17f66a98d/crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs#L1551-L1610

We are handling `Type::Lit` (at line 1557), but we have to handle `Type:::Tpl`, too.

While handling `Type::Tpl`, we should return a new `Type::Tpl`. The expressions of `Type::Tpl` can be copied for string operations.

ever0de:
**Description:**
Handles when tpl comes into the `intrinsics` function.

**BREAKING CHANGE:**
What was reported as an error is changed to normal operation.

fixed #210


crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1537,19 +1537,71 @@ impl Analyzer<'_, '_> {
...
+                        } else {
+                            let mut res = String::new();
+                            let mut chars = s.value.chars();
+
+                            res.extend(chars.next().into_iter().flat_map(|v| v.to_lowercase()));
+                            res.push_str(chars.as_str());
+
+                            res
+                        }
+                    }
```
ever0de, line 1570: Can you turn this into a function?
I think it can be handled by changing `JsKeyword` to Atom.

kdy1: I think so. Do you want me to do it?

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1537,19 +1537,71 @@ impl Analyzer<'_, '_> {
     pub(crate) fn expand_intrinsic_types(&mut self, span: Span, ty: &Intrinsic) -> VResult<Type> {
         let arg = &ty.type_args;
 
-        match ty.kind {
-            IntrinsicKind::Uppercase | IntrinsicKind::Lowercase | IntrinsicKind::Capitalize | IntrinsicKind::Uncapitalize => {
```
ever0de, line 1541: Since there is no `non_exhaustive` attribute, the first match statement may not be necessary. What do you think?
Or if this PR doesn't fix it, I'll roll it back.
Or attach the `non_exhaustive` macro to `IntrinsicKind` and express it as an error?

crates/stc_ts_file_analyzer/tests/pass-only/conformance/types/tpl/stringMappingOverPatternLiterals/1.ts
```diff
@@ -0,0 +1,15 @@
...
+// templated
+type ATemplate = `aA${string}`;
+type BTemplate = Uppercase<ATemplate>;
+type CTemplate = Lowercase<ATemplate>;
+
+let aTemp: ATemplate = "aAFoo";
+let bTemp: BTemplate = "AAFOO";
+let cTemp: CTemplate = "aafoo";
+
+export {};
```
ever0de, line 15: Moved to `pass/*` as none of the cases failed.

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
ever0de: I'll try. Can I put it on this PR?

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
kdy1: I think `#[non_exhaustive]` is not required

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
kdy1: Yes, thank you!

ever0de: https://github.com/dudykr/stc/pull/295/commits/3b25c0af65f0e274264f3a4b3c0c6bb7c8cc4725
I made
Is `JsWord` <-> `Atom` conversion correct?

crates/stc_ts_file_analyzer/tests/pass-only/conformance/types/tpl/stringMappingOverPatternLiterals/1.ts
kdy1: Please move it back to `pass/`. 
`pass-only` does not dump types

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1869,3 +1858,37 @@ pub(crate) fn left_of_expr(t: &RExpr) -> Option<&RIdent> {
         _ => None,
     }
 }
+
+fn apply_intrinsics(intrinsics: &IntrinsicKind, raw: &Atom) -> Atom {
+    match intrinsics {
+        IntrinsicKind::Uppercase => raw.to_uppercase(),
```
kdy1, line 1864: Please use `.to_ascii_uppercase` instead

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1869,3 +1858,37 @@ pub(crate) fn left_of_expr(t: &RExpr) -> Option<&RIdent> {
...
+        IntrinsicKind::Lowercase => raw.to_lowercase(),
```
kdy1, line 1865: Same here. Please use ascii method instead

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1869,3 +1858,37 @@ pub(crate) fn left_of_expr(t: &RExpr) -> Option<&RIdent> {
...
+        IntrinsicKind::Capitalize => {
+            if raw.is_empty() {
+                "".into()
+            } else {
+                let mut res = String::new();
+                let mut chars = raw.chars();
+
+                res.extend(chars.next().into_iter().flat_map(|v| v.to_uppercase()));
```
kdy1, line 1873: Same here, and we don't need iterator here

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1869,3 +1858,37 @@ pub(crate) fn left_of_expr(t: &RExpr) -> Option<&RIdent> {
...
+            }
+        }
+        IntrinsicKind::Uncapitalize => {
+            if raw.is_empty() {
+                "".into()
+            } else {
+                let mut res = String::new();
+                let mut chars = raw.chars();
+
+                res.extend(chars.next().into_iter().flat_map(|v| v.to_lowercase()));
```
kdy1, line 1886: Same here

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1537,58 +1537,47 @@ impl Analyzer<'_, '_> {
...
-                                    let mut chars = s.value.chars();
-
-                                    res.extend(chars.next().into_iter().flat_map(|v| v.to_lowercase()));
-                                    res.push_str(chars.as_str());
-
-                                    res
-                                }
-                            }
-                        };
+        match self.normalize(None, Cow::Borrowed(&arg.params[0]), Default::default())?.as_ref() {
```
kdy1, line 1540: Please use `.normalize()` instead of `as_ref()`

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1537,58 +1537,47 @@ impl Analyzer<'_, '_> {
...
+            Type::Lit(LitType { lit: RTsLit::Str(s), .. }) => {
+                let new_val = apply_intrinsics(&ty.kind, &Atom::new(s.value.as_ref()));
```
kdy1, line 1542: Please use normal reference/dereference instead of `as_ref()`

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1537,58 +1537,47 @@ impl Analyzer<'_, '_> {
...
+
+                return Ok(Type::Lit(LitType {
+                    span: arg.params[0].span(),
+                    lit: RTsLit::Str(RStr {
+                        span: arg.params[0].span(),
+                        value: swc_atoms::JsWord::from(new_val.as_ref()),
```
kdy1, line 1548: Same here. Please `JsWord::from` instead

kdy1 (changes requested):

crates/stc_ts_file_analyzer/tests/pass-only/conformance/types/tpl/stringMappingOverPatternLiterals/1.ts
ever0de: If `std-err` doesn't print, should I upload an empty file?

kdy1: Yes, but no file is also fine

ever0de:
It makes `apply_intrinsics` somewhere I don't know where and `check.sh` fails.
Can I roll back once? (keep testcases in `pass` path)

kdy1:
Feel free to do anything while it's a draft. I'll request change once the PR is ready for review

ever0de:
Let's return the CI per change in the current state.
I think it's probably failing by changing as_ref to normalize.

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
ever0de: Sorry, I didn't understand exactly, 
isn't it necessary?
To make only the first letter Uppercase, there may be a way to import it as a mutable reference and modify it like this: `.get_mut(0..1).map(|prefix| prefix.to_uppercase())`.

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
ever0de: Using `normalize` will fail, as it gets to the branch using `assign_to_intrinsic`.
Is fixing this included in the issue? Or, is it better to leave it as `as_ref()`?

kdy1: I mean `Type.normalize()`

crates/stc_ts_file_analyzer/tests/pass/types/tpl/stringMappingOverPatternLiterals/1.ts
```diff
@@ -1,11 +1,15 @@
...
 type ATemplate = `aA${string}`;
 type BTemplate = Uppercase<ATemplate>;
+type CTemplate = Lowercase<ATemplate>;
 
-var b: BTemplate
+let aTemp: ATemplate = "aAFoo";
+let bTemp: BTemplate = "AAFOO";
+let cTemp: CTemplate = "aafoo";
 
-b = 'AAFOO'
```
kdy1, line 9: Please restore this line

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
kdy1: It's not necessary because it's an ASCII operation.
But my bad, I misread the variant.

Can you pre-allocate string above, and use `to_asscii_uppercase`?

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
kdy1: `apply_intrinsics` should accept `&str`

kdy1 (approved):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
ever0de: Looks like you are already assigning to `res` https://github.com/dudykr/stc/pull/295/commits/33b388860ccf2baf95bf70e1f999217e5bc190a1
Is it right to want a change like this?

kdy1: Yes

ever0de:
@kdy1 
It looks like all of the 'good-first-issues' currently assigned are in progress, are there any additional issues I can work on?
https://github.com/dudykr/stc/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

kdy1:
Seems like some issues are not assigned to someone, e.g. https://github.com/dudykr/stc/issues/186 or https://github.com/dudykr/stc/issues/176
I'll try creating more good-first issues

Issue #181, opened 2022-11-03
Allow assignment of tuples with different length if it's `undefined` or `any`


kdy1:
https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_type_checker/tests/conformance/es6/destructuring/destructuringArrayBindingPatternAndAssignment3.ts#L9-L10

This line should not report an argumentnt error because fields are optional.

<img width="821" alt="image" src="https://user-images.githubusercontent.com/29931815/199725700-b951ae74-a89b-4d3c-bbf6-17828a638fad.png">

The type of arguments are correctly inferred as `[any, any, any, any]` but we should allow assignment of `[1]` to `[any, any, any, any]`.

https://github.com/dudykr/stc/blob/e59631e58fb04578142befb6e5dbc7ef1d52f9db/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs#L1999-L2010

The assignment logic for tuple should be aware of `any` or `undefined` and allow the length to mismatch for these cases.

Beraliv:
Hey @kdy1 !

Can I take this one? I'd like to work on it

kdy1:
Of course, thank you!

Beraliv:
Thanks to your comment @hyp3rflow (link 🔗 – https://github.com/dudykr/stc/issues/179#issuecomment-1302211895).

I've also decided to double check error list in TypeScript repo and it's exactly the same (conformance tests are also identical):

* TypeScript – [destructuringArrayBindingPatternAndAssignment3.errors.txt](https://github.com/microsoft/TypeScript/blob/main/tests/baselines/reference/destructuringArrayBindingPatternAndAssignment3.errors.txt)
* stc - [destructuringArrayBindingPatternAndAssignment3.errors.json](https://github.com/dudykr/stc/blob/main/crates/stc_ts_type_checker/tests/conformance/es6/destructuring/destructuringArrayBindingPatternAndAssignment3.errors.json)

So tests are up-to-date

Beraliv:
Now I'm trying to understand how I can run one specific test and get the same output that you shared on the screenshot

### Working with output

I've checked the typical flow in [CONTRIBUTING.md](https://github.com/dudykr/stc/blob/main/CONTRIBUTING.md#typical-workflow), but it says:

[...]

PR #298, opened 2022-11-15
chore(ts/checker): Enable destructuringArrayBindingPatternAndAssignment3

Beraliv:
**Description:**

```gherkin
Given There are 2 tuples
 When lhs is tuple with `any` elements
  And rhs is tuple with different types
  And lhs is of higher length
 Then rhs is assignable to lhs
```

I simply ignore `AssignFailedBecauseTupleLengthDiffers` when there is a tuple with `any` elements for lhs.

The only regression is `conformance::expressions::arrayLiterals::arrayLiterals3.ts` where `var a0: [any, any, any] = []` is PASSED without errors now (which is wrong).

Can you please recommend a way to distinguish tuple in parameters and tuple in variables?

But we will need to distinguish these cases in the future for sure:

```ts
var a0: [any, any, any] = [];  // ❌ error as tuple of length 0 is NOT assignable to tuple of length 3
var a1: [any?, any?, any?] = []; // ✅ ok, tuple of length 0 is assignable to tuple of length 0 up to 3

(function ([f, g, h, i]) { // ✅ ok, as [1] is number[] by default so here we have f: number, g: number; h: number; i: number
})([1]);
(function ([f, g = 1, h = 2, i = 3]) { // ✅ ok, same but with default parameters
})([1]);
```

Playground 🔗 – https://tsplay.dev/w8xO9w

**BREAKING CHANGE:**

**Related issue (if exists):** https://github.com/dudykr/stc/issues/181


crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2067,7 +2067,21 @@ impl Analyzer<'_, '_> {
...
+                                .iter()
+                                .enumerate()
+                                .filter(|&(i, right_element)| match *(&elems[i]).ty {
+                                    Type::Keyword(KeywordType {
+                                        kind: TsKeywordTypeKind::TsAnyKeyword,
+                                        ..
+                                    }) => true,
+                                    _ => false,
+                                })
+                                .collect::<Vec<_>>();
```
Beraliv, line 2080: this assumption is still not true as:

1. `[1]` is not assignable to`[any, string, string, string]` and should throw an error
2. `[1]` is assignable `[any, any]` and no error should be thrown

So let me do the opposite (with `elems`) to prove it

Beraliv: I got rid of it

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2067,6 +2067,18 @@ impl Analyzer<'_, '_> {
                         // TODO: Handle Type::Rest
 
                         if elems.len() > rhs_elems.len() {
+                            let is_lhs_any_tuple = elems.iter().all(|l| match *l.ty {
```
kdy1, line 2070: Is this correct? I think we should leverage the optional type as you said in https://github.com/dudykr/stc/issues/181#issuecomment-1306378999

kdy1 (approved):

PR #395, opened 2022-12-15
Update(ts/analzer): line_col should display end_column

sunrabbit123:
**BREAKING CHANGE:**

Until now, when we look at the log, we have identified the location of the ts file.
However, only the column start position for that location was received, and the end position could not be obtained.

So I modified that part.

```rs
fn line_col(&self, span: Span) -> String {
    if span.is_dummy() {
        return "".into();
    }
    let loc = self.cm.lookup_char_pos(span.lo);
    let hic = self.cm.lookup_char_pos(span.hi);
    format!("({}:{}-{:?})", loc.line, loc.col_display + 1, hic.col_display + 1)
}
```

None


kdy1 (approved):

sunrabbit123:
image
result display

PR #402, opened 2022-12-15
feat(ts/analyzer): Pass type annotations to non-generic arguments

kdy1:
**Description:**

This is required to infer types correctly in some cases. For example, we can't infer the type of `rest` in https://github.com/dudykr/stc/blob/1f762d8b2becb5a43491d301bf2409615d328112/crates/stc_ts_type_checker/tests/conformance/types/rest/objectRestParameterES5.ts#L5-L6 if we don't pass type annotations.


kdy1:
I filed https://github.com/dudykr/stc/issues/405 because of regressions

PR #406, opened 2022-12-16
fix(ts/analyzer): Fix class property access

CarrotzRule123:
**Description:**
```typescript
class A {
   private foo: number,
   bar: number,
}
new A().bar
```

Causes panic in `tests/conformance/classes/members/privateNames/privateNamesInGenericClasses.ts`
due to `not implemented: access to type elements using private name`


crates/stc_ts_type_checker/tests/conformance/classes/members/privateNames/privateNamesInGenericClasses.ts.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 0,
-    matched_error: 5,
-    extra_error: 9,
-    panic: 0,
+    required_error: 5,
+    matched_error: 0,
+    extra_error: 0,
+    panic: 1,
```
kdy1, line 5: Can you investigate this?

kdy1 (changes requested):
Thanks!

crates/stc_ts_type_checker/tests/conformance/classes/members/privateNames/privateNamesInGenericClasses.ts.stats.rust-debug
CarrotzRule123: The panic is caused by this line https://github.com/dudykr/stc/blob/a0df88a1cfaeccdf444c87b2a70b20a72301707d/crates/stc_ts_types/src/lib.rs#L449. How do I show that a `Type` is private? Can I add something like `is_private` to the type metadata?

kdy1: You should check for `Key::Private` before calling `key.ty()`

kdy1:
Can you rebase?

crates/stc_ts_type_checker/tests/conformance/classes/members/privateNames/privateNameAndObjectRestSpread.ts.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
     required_error: 4,
     matched_error: 0,
-    extra_error: 3,
-    panic: 0,
+    extra_error: 0,
+    panic: 1,
```
kdy1, line 5: Oh, nvm, I think this should be fixed first

PR #410, opened 2022-12-17
fix(ts/analyzer): prevent extra error from being reported in report_error_for_conflicting_parents

togami2864:
**Description:**
Add match arm to prevent `extra_error`.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #412, opened 2022-12-17
fix(ts/analyzer): Top-level throws return `never`

Desdaemon:
**Description:**

Fixes type inference in this case:
```js
function(): number {
  throw undefined
}
```

Originally, `throw` statements do not contribute to return values, and hence fails type inference. Only top-level throws are considered for now.

**BREAKING CHANGE:**

None known yet

None


kdy1 (dismissed):
Thank you!

kdy1:
Can you rebase?

kdy1 (approved):
Thanks!

PR #413, opened 2022-12-18
Make tuple and custom array declared as interface compatible

### LINKED ISSUE #216: Make tuple and custom array declared as interface compatible
# Problem

https://github.com/dudykr/stc/blob/56ae129a0674e81af048e2281653363876ccb11d/crates/stc_ts_type_checker/tests/conformance/types/tuple/arityAndOrderCompatibility01.ts#L1-L5

This type is treated almost as a tuple by `tsc`, but `stc` fails to verify it. 
It's because it extends `Array`.

image

# Solution

I think we should handle it from `Analyzer.normalize`.
Implementing special rules to each function for handling such edge looks like overengineering.

We can add logic for checking if the `ty` is an interface which extends an array-like builtin interface at https://github.com/dudykr/stc/blob/56ae129a0674e81af048e2281653363876ccb11d/crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs#L83-L99

devgony:
**Description:** Make tuple and custom array declared as interface compatible

**BREAKING CHANGE:**

**Related issue (if exists):** To resolve #216


crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -107,6 +107,47 @@ impl Analyzer<'_, '_> {
...
+                type_params,
+                extends,
+                body,
+                metadata,
+            }) => {
+                for extend in extends {
+                    for type_arg in extend.type_args.as_ref() {
+                        for param in type_arg.to_owned().params {
+                            match param {
+                                Type::Array(array) => return Ok(Cow::Owned(Type::Array(array))),
```
devgony, line 122: I tried early return here to see what makes difference but it does not affect the result of `./scripts/test.sh arityAndOrderCompatibility` at all

is `./scripts/test.sh arityAndOrderCompatibility` correct to test?

kdy1: You should add `Type::Interface` to https://github.com/dudykr/stc/blob/d6dc481fd29c9e6e93e268dbd4f11a7b3883a6d2/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs#L512

kdy1: The command itself is correct

devgony:
```
$ ./scripts/test.sh arityAndOrderCompatibility

2 unmatched errors out of 17 errors. Got 5 extra errors.
Wanted: [RefError { line: 19, column: 5, code: "TS2741" }, RefError { line: 22, column: 5, code: "TS2741" }]
Unwanted: [(19, "TS2322"), (22, "TS2322"), (30, "TS2322"), (34, "TS2322"), (35, "TS2322")]
```

I made it to remove `(16, "TS2461")`.
Should i remove all the other 5 extra errors In this issue? 😢

devgony:
```
$ ./scripts/test.sh arityAndOrderCompatibility

2 unmatched errors out of 17 errors. Got 5 extra errors.
Wanted: [RefError { line: 19, column: 5, code: "TS2741" }, RefError { line: 22, column: 5, code: "TS2741" }]
Unwanted: [(19, "TS2322"), (22, "TS2322"), (30, "TS2322"), (34, "TS2322"), (35, "TS2322")]
```

I made it to remove `(16, "TS2461")`.
Should i remove all the other 5 extra errors In this issue? 😢

kdy1:
No, it's okay to split PRs. I'll review this PR once you undraft it.

devgony:
I'm trying to solve [nested_assignment error](https://github.com/dudykr/stc/actions/runs/3729484189/jobs/6325442576) in advance of undraft but cannot get which part is wrong yet.

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -107,6 +107,34 @@ impl Analyzer<'_, '_> {
...
+                span,
+                name,
+                type_params,
+                extends,
+                body,
+                metadata: InterfaceMetadata { common },
+            }) => {
+                for extend in extends {
+                    let types = extend.to_owned().type_args.into_iter().flat_map(|cur| cur.params);
+                    let union_type = Type::new_union(*span, types);
```
devgony, line 120: It almost solved nest_assignment errors but not RegExpExecArray yet.
Could you give me any hint where to get RegExpExecArray type to return Array?

kdy1: `RegExpExecArray` is declared as

```ts
interface RegExpExecArray extends Array<string> {
    index: number;
    input: string;
}
```

So you need to check the type arguments of `Array`

kdy1:
Can you sign the CLA?

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -117,6 +116,36 @@ impl Analyzer<'_, '_> {
...
+                type_params,
+                extends,
+                body,
+                metadata: InterfaceMetadata { common },
+                tracker,
+            }) => {
+                for extend in extends {
+                    let types = extend.to_owned().type_args.into_iter().flat_map(|cur| cur.params);
+                    let union_type = Type::new_union(*span, types);
+                    if let box RExpr::Ident(ident) = &extend.expr {
```
kdy1, line 131: I think we should check for `Type::Array` after using `self.type_of_ts_entity_name`

kdy1 (changes requested):
Thanks!

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -117,6 +116,22 @@ impl Analyzer<'_, '_> {
             | Type::EnumVariant(..)
             | Type::Param(_)
             | Type::Module(_) => return Ok(ty),
+            Type::Interface(Interface {
```
kdy1, line 119: I think this approach is wrong

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -117,6 +116,22 @@ impl Analyzer<'_, '_> {
...
+                span,
+                name,
+                type_params,
+                extends,
+                body,
+                metadata: InterfaceMetadata { common },
+                tracker,
+            }) => {
+                for parent in extends {
+                    let parent = self.type_of_ts_entity_name(parent.span(), &parent.expr, parent.type_args.as_deref())?;
```
kdy1, line 129: A user can add methods to their own array type.

kdy1: And this does not handle tuple types correctly

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
devgony: Did you mean we should add more condition here to decrease the range of match arm?
Or It is totally wrong location to implement?

devgony: Can you give any tips like where to start first?

kdy1: I think some match expressions should be added to https://github.com/dudykr/stc/blob/4033e8b3a1803640fdcf409442a3cc91b900f237/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs#L1036

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -1421,6 +1420,22 @@ impl Analyzer<'_, '_> {
...
+            Type::Interface(Interface {
+                span,
+                name,
+                type_params,
+                extends,
+                body,
+                metadata: InterfaceMetadata { common },
+                tracker,
+            }) => {
+                for parent in extends {
```
kdy1, line 1432: I think this logic is too naive

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -1421,6 +1420,22 @@ impl Analyzer<'_, '_> {
...
+                    let ty = self.type_of_ts_entity_name(parent.span(), &parent.expr, parent.type_args.as_deref())?;
+                    if let Type::Array(_) = &ty {
```
kdy1, line 1434: Checking for `Type::Array` is not enough, and we should handle this by accessing iterating over properties

kdy1:
```ts
interface MyTuple extends Array<unknown> {
    0: number
    1: number
}

declare let a: MyTuple
declare let b: [number, number]

a = b // ok
b = a // error
```

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgLIE8AqBXADgGxQgA9IQATAZ2QEEoo50AebEAaxAHsB3EAPmQBvAFDIxyAAwAuZCGwBbAEbRR4gIwy5SlQF9hw8hAT44UFITDI4MjDgIQDRk2eQXkimQG0tyqABpZBV8AXX04ZABed2QAehjkTjZhRUirWPjoKE4oIA

kdy1:
```ts
interface MyTuple extends Array<any> {
    0: number
    1: number
}

declare let a: MyTuple
declare let b: [number, number]

a = b // ok
b = a // error
```

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgLIE8AqBXADgGxQgA9IQATAZ2QEEoo50AeOEdAPmQG8AoZf5AAYAXMhDYAtgCNofAQEZR46bIC+PHuQgJ8cKCkJhkcURhwEIm7bv3JDyKaIDaymVAA0YyW4C6GuMgAvA7IAPShyAD2ANY8UkHGYRHQUJFQQA

kdy1: As you can see, we can't rely on the type argument in `extends Array<T>`

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -1031,6 +1031,56 @@ impl Analyzer<'_, '_> {
...
+                            tracker,
+                        }) = elem_type.union_type()
+                        {
+                            for (lhs, ty) in lhs.clone().zip(types) {
+                                if let (Some(KeywordType { kind: lkind, .. }), Some(KeywordType { kind: rkind, .. })) =
+                                    (lhs.to_owned().keyword(), ty.keyword())
+                                {
+                                    println!("@@@\n{lkind:#?}:{rkind:#?}");
+                                    let diff = lkind == rkind;
+                                    println!("@@@\n{diff}");
```
devgony, line 1077: 1. Is there any existing function to compare two KeywordType and return reasonable error?
2. Do you think I'm going correct way?

kdy1: 1. You can use `assign_with_opts(...)?` to do it
2. I think you should not reimplement whole logic here, but delegate to a recursive call to `assign_with_opts`

devgony: It looks like we should keep logic here.
Otherwise, it breaks other tests
https://github.com/dudykr/stc/actions/runs/4110781244/jobs/7093929846

kdy1:
I'm not sure if this code is working as there's no stat change. I'll run stat check locally..

kdy1 (changes requested):
CI failed

sunrabbit123:
Is this PR resolved?

I'd like to see some test code added for those entering later.

It would be nice to know right away what you've done.

@devgony @kdy1

kdy1:
@sunrabbit123  I think this PR is far from ready

sunrabbit123:
Sorry about that. I thought the commit was old and being neglected.

devgony:
@sunrabbit123
I haven't added any test cases so far.
Yes.. this PR is still ongoing yet.

devgony:
It seems that this PR is solved by another PR.
Unlike before, `./scripts/test.sh arityAndOrderCompatibility` passes all.
Or should I change or add more test cases ?

kdy1:
Adding more tests would be awsome. Thank you!

kdy1:
Adding more tests would be awesome. Thank you!

crates/stc_ts_type_checker/tests/conformance/types/tuple/arityAndOrderCompatibility01.ts
```diff
@@ -33,3 +33,10 @@ var n3: [number, string] = z;
 var o1: [string, number] = x;
 var o2: [string, number] = y;
 var o3: [string, number] = y;
+
+x = y;
+// x = z; // should get TS2322 but pass
+y = x;
+y = z;
+// z = x; // should pass but got TS2322
```
devgony, line 41: According to TS Playground,
It seems that `x = z` and `z = x` work in opposite directions.
Should I follow TS playground and fix it in this PR?
image

sunrabbit123: I think that this issue is related to the following issue @kdy1 
#1005

crates/stc_ts_type_checker/tests/tsc-stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 3535,
-    matched_error: 6500,
-    extra_error: 771,
-    panic: 74,
+    required_error: 874,
+    matched_error: 1749,
+    extra_error: 284,
+    panic: 20,
```
kdy1, line 5: Can you run check.sh agian?

crates/stc_ts_type_checker/tests/conformance/types/tuple/arityAndOrderCompatibility01.ts
devgony: @sunrabbit123
Thanks for the information.
So it looks like it does not need to be fixed, at least in this PR.

crates/stc_ts_type_checker/tests/tsc-stats.rust-debug
devgony: I got `thread 'conformance::types::uniqueSymbol::uniqueSymbolsDeclarations.ts' has overflowed its stack`
Should i pass some option to avoid the overflow?

sunrabbit123: The way is simple

1. Check that the code you wrote affects stackoverflow.
2. If it does, insert some stackguards.
3. if it doesn't, it's broken and you can report it in an issue

Thanks you

crates/stc_ts_type_checker/tests/conformance/types/tuple/arityAndOrderCompatibility01.ts
```diff
@@ -33,3 +33,8 @@ var n3: [number, string] = z;
 var o1: [string, number] = x;
 var o2: [string, number] = y;
 var o3: [string, number] = y;
+
+x = y;
```
kdy1: Please revert this

kdy1: This test file is copied from the official tsc conformance test suite, so we should not modify it.

kdy1 (changes requested):

kdy1 (commented):
It seems like this PR does not improve the test stats?

PR #416, opened 2022-12-18
fix(ts/analyzer): Fix module declaration evaluation order

Desdaemon:
**Description:**

Fixes the following example:
```js
class Foo {}
var foo: Foo
module Foo {
  var inside = foo // 'foo' is not visible
}
```

While generating the dependency graph, `module Foo` used to be declared both as a type and a variable. This causes the module to be evaluated before the var-declaration, therefore making `foo` invisible to the module. However, it is [not currently valid to refer to modules in type annotations](https://www.typescriptlang.org/play?#code/LYewJgrgNgpgBAMRCOBvAUHLcYA8AOIATgC5wkCe+8SIAyiUQJYB2A5nALxwDOjrbdAF90ANwCGROADNkALkTI4AemVwWIMhKhMwYyTL7N2C2gDpaDYx1VwmJHjgLEHcSTAA0cABYgA7jCiMERAA), so they shouldn't be declared as one.

**BREAKING CHANGE:**

None known yet

Partially addresses #361

### DIFF
--- a/crates/stc_ts_file_analyzer/tests/pass/es6/module/conflictsWithClass/1.swc-stderr
+++ b/crates/stc_ts_file_analyzer/tests/pass/es6/module/conflictsWithClass/1.swc-stderr
@@ -1,0 +1,27 @@
+
+  x Type
+   ,-[$DIR/tests/pass/es6/module/conflictsWithClass/1.ts:8:3]
+ 8 | var inside = foo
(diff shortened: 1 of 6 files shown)


kdy1 (approved):
Thank you!
I rebased the PR to fix a merge conflict

PR #417, opened 2022-12-19
fix(ts/analyzer): add the check of the type args to satisfy constraints

togami2864:
**Description:**
- added TS2334
- Implemented a check on the constraints in a class with generic types to ensure that the class only accepts valid input.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

PR #429, opened 2022-12-20
fix(ts/analyzer): allow assigning to a mapped type that takes itself as arg

togami2864:
**Description:**
handle these cases.
```
function f1<T>(x: Required<T>, y: T) {
    y = x
    x = y // error because `Require` has -? modifiers
}

function f2<T>(x: Partial<T>, y: T) {
    y = x; // error because `Partial` has +? or ? modifiers
    x = y;
}
```

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #434, opened 2022-12-22
feat(ts/analyzer): Improve generic inference

kdy1:
**Description:**

 - Fix logic for type inference.
 - Remove `InferredType`.
 - Fix the inference of enum definitions.
 - Handle the assignment of a read-only array to a tuple.
 - Ignore the enum variant name and the tuple length while checking compatibility.


kdy1 (commented):
I'll merge this PR to split PRs

Issue #431, opened 2022-12-21
If an overflow occurs, the result should be any.


sunrabbit123:
```ts
type NonOptionalKeys<T> = { [P in keyof T]: undefined extends T[P] ? never : P }[keyof T];
type Child<T> = { [P in NonOptionalKeys<T>]: T[P] }

export interface ListWidget {
    "type": "list",
    "minimum_count": number,
    "maximum_count": number,
    "collapsable"?: boolean, //default to false, means all expanded
    "each": Child<ListWidget>;
}

type ListChild = Child<ListWidget>
declare let x: ListChild;
x.type;
```
x should be any, Because A causes overflow.

sunrabbit123:
[checker.ts](https://raw.githubusercontent.com/microsoft/TypeScript/main/src/compiler/checker.ts)
search `isDeeplyNestedType`

PR #437, opened 2022-12-23
Update union ranking

sunrabbit123:
**Description:**

Add Union Type priority rule


crates/stc_ts_file_analyzer/tests/visualize/types/typeLit/1.swc-stderr
```diff
@@ -30,4 +30,4 @@ Error:
    `----
 
 Error: 
-  > (string | null)
+  > string
```
kdy1, line 33: I think this is wrong

kdy1 (changes requested):
Thanks!
I found some regressions

crates/stc_ts_file_analyzer/tests/visualize/types/typeLit/1.swc-stderr
sunrabbit123: because option is not apply
This is why #436 should be passed.

sunrabbit123:
I think that the panic arising from IndexAccessType.prop_ty.index_type.normalize match is a necessary evil.
because There is no countermeasure against that the panic arising from overflow

kdy1:
You can use `stack::track(span)?` instead

sunrabbit123:
but `stack::track(span)?` call `stackOverFlow` 
This doesn't change anything.

My problem is just 'check'.sh` stops midway due to stack overflow.

kdy1:
You can do like

```rust
let _guard = match stack::track(span) {
    Ok(v) => v,
    Err(_) => return ty,
};
```

kdy1 (changes requested):
Please revert base unit test files

sunrabbit123:
Is this what you said?

kdy1:
No, I meant tests in stc_ts_file_analyzer

sunrabbit123:
Then is this it?

kdy1:
Yes, thanks!

crates/stc_ts_type_checker/tests/tsc-stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 4245,
-    matched_error: 5639,
-    extra_error: 743,
-    panic: 72,
+    required_error: 4248,
+    matched_error: 5636,
+    extra_error: 720,
+    panic: 75,
```
kdy1, line 5: Please fix this

kdy1 (changes requested):

crates/stc_ts_type_checker/tests/tsc-stats.rust-debug
sunrabbit123: Are you talking about the overflow due to circulation reference?

The next method is not a solution
```rs
let _guard = match stack::track(span) {
    Ok(v) => v,
    Err(_) => return ty,
};
```

It's just a stopgap
What's the difference from a step back for this step forward?

but, I will try to reduce the other panic, but there is nothing I can do about panic to prevent overflow

kdy1: Why is stopgap bad?
See [checker.ts](https://raw.githubusercontent.com/microsoft/TypeScript/main/src/compiler/checker.ts) and search for `isDeeplyNestedType`

sunrabbit123: I think the information should be mentioned here(#431) and worked on another PR.  

If you use the following code, it can cause an unexpected error.
```rs
let _guard = match stack::track(span) {
    Ok(v) => v,
    Err(_) => return ty,
};
```
causing inaccurate type inference.

However, to use a method to infer all overflows as cyclic reference errors,
This also allows us to infer all pure overflows by circular reference.
```rs
let _guard = match stack::track(span) {
    Ok(v) => v,
    Err(_) => return ErrorKind::CircularRefrenceError,
};
```
Then, as a temporary measure, you can insert the code
The filter of "If IndexAccess Types" that I put in as a temporary measure before,
If they're both temporary, they're both forced to wait for #431.

sunrabbit123: The way to reduce regression right now is to insert a circular reference error instead of an Overflow error.
Even now, most overflows are caused by cyclical reference.

sunrabbit123: or insert next code
```rs
Type::Union(..) => {
                                        let some = prop_ty.index_type.normalize();
                                        dbg!(dump_type_as_string(&some));
                                        dbg!(&some);
                                        dbg!(self.rule().strict_null_checks);
                                        panic!("{:?}", prop_ty);
                                    }
                                    _ => {
                                       
                                    }
```

sunrabbit123: umm.... I think that this overflow isn't throw `stack::trace`

next is error message
```
thread 'pass_tests__pass__types__mapped__recursiveMappedTypes__1_ts' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass `--test base`
```

but, I know that the overflowError thrown by stack::trace has a different message.

I don't know how to solve this problem.
The method you suggested is meaningful when overflow is by stack::trace.
The method I suggested was also,

sunrabbit123:
My limit is that there are two increased panics in total.

sunrabbit123:
I'm finally back.

crates/stc_ts_file_analyzer/tests/visualize/exprs/bin/comparison/null-1.swc-stderr
```diff
@@ -69,7 +69,7 @@ Error:
     `----
 
 Error: 
-  > (string | undefined)
+  > string
```
kdy1, line 72: I think this is a wrong change.

kdy1: Why did you remove `undefined` from the union?

kdy1 (changes requested):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -359,6 +359,27 @@ impl Analyzer<'_, '_> {
...
+                                    || t.is_kwd(TsKeywordTypeKind::TsBigIntKeyword)
+                                    || t.is_kwd(TsKeywordTypeKind::TsSymbolKeyword)
+                                    || t.is_kwd(TsKeywordTypeKind::TsObjectKeyword)
+                                    || t.is_kwd(TsKeywordTypeKind::TsVoidKeyword)
+                                    || t.is_kwd(TsKeywordTypeKind::TsIntrinsicKeyword)
+                                    || t.is_type_lit()
+                                    || t.is_instance()
+                                    || t.is_interface()
+                            }) {
+                                if !self.rule().strict_null_checks {
```
kdy1, line 379: This is too dangerous. 
There are some cases where we have to preserve the list of types in a union or an intersection

crates/stc_ts_file_analyzer/tests/visualize/exprs/bin/comparison/null-1.swc-stderr
sunrabbit123: He is not strict

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
sunrabbit123: I think you can use the preserve option.

kdy1: This behavior should be opt-in

sunrabbit123: I think too, But I've never seen such a situation. Is there an example of ts that fits the following situation?

kdy1: There was, but I didn't write it somewhere and I don't remember

sunrabbit123: I don't understand your opinion

`null` and `undefined` is right that A and B disappear if they are not strict.

I can't think of anything, and if it's a special case,
These are similar cases to TypeParam and Condition.
If there's a problem then,
Someone can add a condition to this part.

If not, please make me understand through ts example.

PR #438, opened 2022-12-26
feat(ts/analyzer): handle a protected/private constructor

togami2864:
**Description:**

- Show TS2673 if calling a private constructor. 
- Show TS2674 if calling a protected constructor outside the subclass. 
- Show TS2675 if extending a class that has a private constructor.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #442, opened 2022-12-28
fix(ts/analyzer): show TS2340 when accessing to super class property via super keyword

togami2864:
**Description:**
Tiny fix about property access via `super` keyword.
`super` can access class method, but TypeScript disallows to  access property.

**BREAKING CHANGE:**


kdy1 (dismissed):
Thank you!

kdy1 (approved):

PR #443, opened 2022-12-28
feat(ts/analyzer): Normalize an enum on RHS

kdy1:
**Description:**

We can normalize an enum if it's the RHS.


kdy1:
Closed as I was wrong

PR #457, opened 2023-01-06
feat: add vscode debugging config

togami2864:
**Description:**
I'm not sure if this is the best approach, but I believe it could be beneficial for new contributors. 

This PR adds a vscode debugging configuration to make it easier to check and learn how to run the type checker while developing. 

Currently, we usually use `./scripts/test.sh` to check a single file, but it takes a lot of time due to regression checking and generates enormous logs. Additionally, it can be difficult to run the type checker with the debugger. 

**BREAKING CHANGE:**

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,7 +24,7 @@ jobs:
 
       - run: yarn
 
-      - run: npx cspell '**/*'
+      - run: npx cspell --no-progress '**/*'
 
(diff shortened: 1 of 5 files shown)


kdy1 (dismissed):
Thank you!

I think this is a good idea. I always wanted to attach debugger, but I just didn't bother to do it.

kdy1 (approved):

PR #465, opened 2023-01-09
fix(ts/analyzer): make unknown assignable to null | undefied | {}

togami2864:
**Description:**
- Fix some panic that occurs when accessing the index type with a TS keyword like `string`
- make unknown assignable to `null | undefied | {}`

I will fix the extra_errors that were introduced by fixing the panic in another PR.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #470, opened 2023-01-09
 feat(ts/analyzer): Do not report CannotAccessPrivatePropertyFromOutside if the property is not actually being accessed from the outside

Subv:
**Description:**

The original code was reporting this error if at least one property member of the class was #private.
Example repro of the problem:

```ts
// @target: esnext
class Class1 {
  #private_value!: number;

  get call_get() {
    return 1;
  }

  set call_set(value: number) {
  }
}

let c1 = new Class1();
c1.call_set = 5;

```

Only report a CannotAccessPrivatePropertyFromOutside when a class property is actually being accessed.

I am currently getting 2 more panics and overflows than what is currently reported in the stats files, namely in these files:

`crates/stc_ts_type_checker/tests/conformance/classes/members/privateNames/privateNamesAndMethods.ts`
`crates/stc_ts_type_checker/tests/conformance/classes/members/privateNames/privateNamesInGenericClasses.ts`

Even a clean clone seems to fail those panics, I included them in the commit for visibility.

My setup:
2021 MacBook Pro M1 Pro (Aarch64)
Node 18.12.1 installed using nvm
rustc 1.66.0-nightly (e7119a030 2022-09-22) (nightly-2022-09-23-aarch64-apple-darwin)


crates/stc_ts_type_checker/tests/conformance/classes/members/privateNames/privateNamesInGenericClasses.ts.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 0,
-    matched_error: 5,
-    extra_error: 9,
-    panic: 0,
+    required_error: 5,
+    matched_error: 0,
+    extra_error: 0,
+    panic: 1,
```
kdy1, line 5: Can you investigate this?

crates/stc_ts_type_checker/tests/conformance/classes/members/privateNames/privateNamesAndMethods.ts.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
     required_error: 6,
     matched_error: 0,
-    extra_error: 7,
-    panic: 0,
+    extra_error: 0,
+    panic: 1,
```
kdy1, line 5: And this?

kdy1 (changes requested):
Thank you!

Subv:
This PR seems to be superseded by https://github.com/dudykr/stc/pull/476 which also fixes other things, I would suggest concentrating the efforts on that one.

As for the panics, they're stack overflows and I'm unsure how to debug them. My theory is that it's something related to my particular setup (Aarch64 MacOS) since they happen even with a clean clone, any pointers on how to debug that are appreciated.

kdy1:
You need a debugger to debug it, and take a look at check.sh or test.sh. After then you will find some env vars

kdy1:
Closing as this is superseded

PR #471, opened 2023-01-10
remove duplicate

sunrabbit123:
**Description:**

sometimes conformance has duplicate values
Maybe it's just me.
It's a function for me.

**BREAKING CHANGE:**

I remove duplicate value with `set`
`set` can't duplicate value

### DIFF
--- a/crates/stc_ts_type_checker/scripts/_/sort.py
+++ b/crates/stc_ts_type_checker/scripts/_/sort.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 import sys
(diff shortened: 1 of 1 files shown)


kdy1:
I think problem is your `uniq`, not `sort.py`

kdy1:
See https://github.com/dudykr/stc/blob/d3d6a8fa98f5d6cceefb046124fcf6b63d9305cb/crates/stc_ts_type_checker/scripts/sort.sh#L5
There's `| uniq |`

sunrabbit123:
I'm so sad
ok, I will close this pr

PR #472, opened 2023-01-10
feat(ts/analyzer): resolve constrolFlowIfStatement.ts

sunrabbit123:
**Description:**

resolve never assign never

**BREAKING CHANGE:**

add next code
```rs
if to.is_kwd(TsKeywordTypeKind::TsNeverKeyword) {
    if let Type::Param(TypeParam { constraint: Some(ty), .. }) = rhs {
        if ty.is_never() {
            return Ok(());
        }
    }
    fail!()
}
```

and 

replace next code
```rs
if u32::from(is_str) + u32::from(is_num) + u32::from(is_bool) >= 2 {
    return Ok(());
}
```

to 

```rs
if u32::from(is_str) + u32::from(is_num) + u32::from(is_bool) >= 2 {
	fail!()
}
```

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
@@ -757,23 +757,6 @@ impl Analyzer<'_, '_> {
             return Ok(());
         }
 
-        if rhs.is_any() {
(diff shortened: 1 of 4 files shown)


kdy1 (approved):
Thank you!

PR #475, opened 2023-01-10
feat: support basic implementation of class static blocks

togami2864:
**Description:**
implemented a simple pattern for static class fields

**BREAKING CHANGE:**


crates/stc_ts_type_checker/tests/conformance/classes/classStaticBlock/classStaticBlock16.ts.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 3,
-    matched_error: 0,
-    extra_error: 2,
+    required_error: 2,
+    matched_error: 1,
+    extra_error: 3,
```
togami2864, line 4: Can I split this `extra_error` into another issues? I know  the cause of the error but it seems difficult to find a good solution and it may become slightly complex. 
In a nutshell, this error occurs when referencing a class within a method or static class field, resulting in a stack overflow error.
```
let getX: (c: C) => number;
class C {
  #x = 1;
  static {
    getX = (obj: C) => obj.#x; // stack overflow
  }
}
```
```
class C {
    method(other: C) {
        const obj = { ...other }; // stack overflow error
    }
}
```

kdy1: Of course, feel free to file an issue and postpone them

togami2864: create as an issue #486

kdy1: Cool, thank you!

togami2864:
And could you teach me  the difference between test cases that has `.ts.stats.rust-debug` and those that do not?
I think there are still cases that need to be handled in the class static  block, but many of those don't have `.ts.stats.rust-debug`.

kdy1:
Oh? I think it's a bug.
I think it's due to an early panic, which occurs while loading the test list.

togami2864:
For example, ./scripts/test.sh [classStaticBlock2](https://github.com/dudykr/stc/blob/main/crates/stc_ts_type_checker/tests/conformance/classes/classStaticBlock/classStaticBlock2.ts) success but there's no `classStaticBlock2.ts.stats.rust-debug`

kdy1:
I found that the test was ignored. I'll work to enable those tests.

kdy1:
https://github.com/dudykr/stc/blob/6582e0c5bad2d9cca2e276b691cae16cd101635e/crates/stc_ts_type_checker/tests/tsc.multiresult.txt#L116

kdy1:
Ah, I ignored those tests because there's no error loading logic for e.g. `classStaticBlock3(target=es2022).errors.json`

kdy1 (approved):
Thank you!

PR #477, opened 2023-01-11
if has error, evaluate never

sunrabbit123:
**Description:**

if code has SwitchCaseTestNotCompatible, This should be inferred as never. 

**BREAKING CHANGE:**

just add branch

```rs
let r = if has_switch_case_test_not_compatible {
    Type::Keyword(KeywordType {
        span,
        kind: TsKeywordTypeKind::TsNeverKeyword,
        metadata: Default::default(),
        tracker: Default::default(),
    })
} else {
    prevent_generalize(&mut r);
    r.make_clone_cheap();
    r
};
```


kdy1 (approved):
Thanks!

kdy1 (approved):

PR #478, opened 2023-01-11
feat(ts/analyzer): Assign empty typeLit

sunrabbit123:
**Description:**

```ts
type Hmm = { a: string } extends {} ? string : never;
```
next type should be string, but this is never at stc


kdy1 (approved):
Thank you!

kdy1 (approved):

PR #484, opened 2023-01-13
feat(ts/analyzer): the right type inference of typeParam with unknown

sunrabbit123:
**Description:**

```ts
//@strict: true
function doSomething1<T extends unknown>(value: T): T {
  if (value === undefined) {
    return value; // value: T extends unknown
  }
  value; // T & ({} | null)
  if (value === 42) {
    value; // T & {}
    throw Error("Meaning of life value");
  }
  value; // value: T & ({} | null)
  throw Error("Meaning of life value");
  // return value; //T & ({} | null)
}

doSomething1(undefined); // undefined
```
this type inference is so terrible

**BREAKING CHANGE:**

First. when add type fact, I append edge case
Second. When while exclude type,  I append edge case of unknown and typeParam


.vscode/settings.json
```diff
@@ -41,5 +42,5 @@
   ],
   "rust-analyzer.checkOnSave.command": "clippy",
   "debug.allowBreakpointsEverywhere": true,
-  "rust-analyzer.debug.openDebugPane": true,
-}
\ No newline at end of file
+  "rust-analyzer.debug.openDebugPane": true
```
kdy1, line 45: Please revert this

sunrabbit123: just remove `,`
is it serious action?

kdy1: Oh I misread it, my bad. Sorry

crates/stc_ts_file_analyzer/tests/pass/types/unknown/controlFlow/1.swc-stderr
```diff
@@ -0,0 +1,128 @@
...
+  | }))
+
+  x Type
+   ,-[$DIR/tests/pass/types/unknown/controlFlow/1.ts:8:3]
+ 8 | if (value === 42) {
+   :     ^^^^^
+   `----
+
+Error: 
+  > never
```
kdy1, line 55: This is wrong

crates/stc_ts_file_analyzer/tests/pass/types/unknown/controlFlow/1.swc-stderr
```diff
@@ -0,0 +1,128 @@
...
+  > boolean
+
+  x Type
+   ,-[$DIR/tests/pass/types/unknown/controlFlow/1.ts:9:5]
+ 9 | value; // T & {}
+   : ^^^^^
+   `----
+
+Error: 
+  > never
```
kdy1, line 73: This is also wrong

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -2412,9 +2437,15 @@ impl Analyzer<'_, '_> {
                 constraint: Some(constraint),
                 ..
             }) => {
-                self.exclude_type(span, constraint, &excluded);
-                if constraint.is_never() {
-                    *ty = Type::never(span, Default::default());
+                if constraint.is_unknown() {
+                    let mut constraint_temp = constraint.normalize().clone();
```
kdy1, line 2441: `normalize().clone()` is anti-pattern

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -2436,6 +2467,41 @@ impl Analyzer<'_, '_> {
...
+                        tracker: Default::default(),
+                    }),
+                    Type::TypeLit(TypeLit {
+                        span: *span,
+                        members: vec![],
+                        metadata: Default::default(),
+                        tracker: Default::default(),
+                    }),
+                ];
+                unknown.retain(|ty| !ty.type_eq(excluded.normalize()));
```
kdy1, line 2499: You don't need to call `Type.normalize()` for `type_eq`

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
sunrabbit123: thx, i fixed it

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
sunrabbit123: thx, i fixed it

crates/stc_ts_file_analyzer/tests/pass/types/unknown/controlFlow/1.swc-stderr
sunrabbit123: yes, this is wrong.

but, There is a reason why this value is never.
This is because find_var gets this value as undefined.
I don't know why I can't find the value right above.

crates/stc_ts_file_analyzer/tests/pass/types/unknown/controlFlow/1.swc-stderr
sunrabbit123: this too

kdy1 (approved):

PR #494, opened 2023-01-15
fix(ts/analyzer): improve type guard of the form `typeof x === s`

togami2864:
**Description:**
from [this test case](https://github.com/dudykr/stc/blob/main/crates/stc_ts_type_checker/tests/conformance/expressions/typeGuards/typeGuardOfFormTypeOfOther.ts)

```
A type guard of the form typeof x === s, where s is a string literal with any value but 'string', 'number' or 'boolean',
  - when true, removes the primitive types string, number, and boolean from the type of x, or
  - when false, has no effect on the type of x.

A type guard of the form typeof x !== s, where s is a string literal,
  - when true, narrows the type of x by typeof x === s when false, or
  - when false, narrows the type of x by typeof x === s when true.
```

**BREAKING CHANGE:**

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/expr/bin.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/expr/bin.rs
@@ -918,47 +918,6 @@ impl Analyzer<'_, '_> {
 
         let c = Comparator { left: l, right: r };
 
-        if let RExpr::Unary(RUnaryExpr {
-            op: op!("typeof"),
-            ref arg,
-            ..
-        }) = l
-        {
-            match r {
-                RExpr::Lit(RLit::Str(RStr { ref value, .. })) => match (TypeFacts::typeof_eq(value), TypeFacts::typeof_neq(value)) {
-                    (None, None) => {
-                        let name = Name::try_from(&**arg).unwrap();
-
-                        self.cur_facts.true_facts.excludes.insert(
-                            name.clone(),
-                            vec![
-                                Type::Keyword(KeywordType {
-                                    span,
-                                    kind: TsKeywordTypeKind::TsStringKeyword,
-                                    metadata: Default::default(),
-                                    tracker: Default::default(),
-                                }),
-                                Type::Keyword(KeywordType {
-                                    span,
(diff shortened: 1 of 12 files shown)


crates/stc_ts_file_analyzer/src/analyzer/expr/bin.rs
```diff
@@ -941,14 +942,82 @@ impl Analyzer<'_, '_> {
...
+                            // A type guard of the form typeof x === s and typeof x !== s,
+                            // where s is a string literal with any value but 'string', 'number' or
+                            // 'boolean'
+                            let name = Name::try_from(&**arg).unwrap();
+
+                            if is_eq {
+                                //  - typeof x === s
+                                //  removes the primitive types string, number, and boolean from
+                                //  the type of x in true facts.
+                                self.cur_facts.true_facts.excludes.insert(
```
kdy1, line 966: Please use `self.cur_facts.true_facts.excludes.entry(name.clone()).or_default().extend` instead

crates/stc_ts_file_analyzer/src/analyzer/expr/bin.rs
```diff
@@ -941,14 +942,82 @@ impl Analyzer<'_, '_> {
...
+                                            metadata: Default::default(),
+                                            tracker: Default::default(),
+                                        }),
+                                    ],
+                                );
+                            } else {
+                                //  - typeof x !== s
+                                //  removes the primitive types string, number, and boolean from
+                                //  the type of x in false facts.
+                                self.cur_facts.false_facts.excludes.insert(
```
kdy1, line 993: Same here.

kdy1 (changes requested):
Thank you!

`insert()` removes the current entry, so we should use `entry().or_default().extend()` instead

kdy1 (dismissed):
Thank you!

kdy1 (dismissed):

kdy1 (approved):

kdy1 (changes requested):
CI failed

kdy1 (approved):

PR #498, opened 2023-01-15
test(ts/base): Cache `tsc` invokation result

kdy1:
**Description:**

This is to make iteration speed faster.


kdy1:
Performance difference: 1 minute => 41 seconds

PR #502, opened 2023-01-15
test(ts/base): Make base test faster

kdy1:
It was slow due to a bug


kdy1:
Performance difference (`check.sh`): 45 seconds => 37 seconds

PR #510, opened 2023-01-15
fix: Fix static property access with a private key

CarrotzRule123:
**Description:**
```javascript
class A {
    static #prop = 1
}

A.#prop; // Error```


kdy1 (dismissed):
Thank you!

PR #512, opened 2023-01-17
Fix(ts/analyzer): TS2708 should only occur on empty modules.

sunrabbit123:
**Description:**

Fix wrong error message about module(2631, 2708)


kdy1 (approved):
Thanks!

PR #516, opened 2023-01-18
fix(ts/analyzer): Change condition for `TS2340`

CarrotzRule123:
**Description:**

Only throw `TS2340` in `ES5`
```typescript
class C {
    static a = 1;
}

class D extends C {
    static b = super.a + 1;
}
```


kdy1 (approved):
Thanks!

PR #517, opened 2023-01-19
feat(ts/analyzer): the right infer about union instance of 

sunrabbit123:
**Description:**

```ts
// a type with a prototype, it overrides the construct signature
interface GConstructor {
  prototype: G1; // high priority
  new (): G2; // low priority
}
interface G1 {
  foo1: number;
}
interface G2 {
  foo2: boolean;
}
declare var G: GConstructor;

var obj13: G1 | G2;
if (obj13 instanceof G) {
  // narrowed to G1. G1 is return type of prototype property.
  obj13.foo1;
}
```
obj13 shoulde infer G1, But it was previously deduced as G2.


kdy1 (dismissed):
Thank you!

Can you rebase?

kdy1 (approved):

PR #518, opened 2023-01-19
feat(ts/analyzer): the right infer about union instance of duplicate constructure ret_ty 

sunrabbit123:
**Description:**

```ts
interface EConstructor {
  new (): E1 | E2;
}

interface A {
  foo: string;
}

interface E1 {
  foo: string;
  bar1: number;
}
interface E2 {
  foo: string;
  bar2: number;
}
declare var E: EConstructor;

var obj9: E1 | A;
if (obj9 instanceof E) {
  obj9.foo;
  obj9.bar1;
}
```

`obj9` should be `E1`


kdy1 (approved):
Thanks!

PR #528, opened 2023-01-20
feat: upgrade builtin types

togami2864:
**Description:**
This PR upgrades built in types in `./crates//stc_ts_builtin_types/lib/*` and enable Es2021 & Es2022.
Got `d.ts` files from [TypeScript main branch(2023/01/20)](https://github.com/microsoft/TypeScript).

**BREAKING CHANGE:**

**Related issue (if exists):**
#467
@kdy1  This PR makes 4 extra panics and I did't enable  `es2022.intl.d.ts` because it causes a lot of regression.
Since the diff is so large, can I split the PR with another issue?

PR #535, opened 2023-01-21
fix(ts/analyzer): resolve some panic

togami2864:
**Description:**
follow-up #528

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #536, opened 2023-01-23
fix: Fix evaluation of `this` and `staticthis`

CarrotzRule123:
**Description:**

Create instance from `StaticThis` not `This`.
Check whether `this` is static beforehand and not during member access.

```typescript
class Base {
    static create() {
        return new this(); // StaticThis not This
    }
}
```


kdy1 (approved):
Thanks!

Can you rebaae?

kdy1 (dismissed):

kdy1 (approved):

PR #544, opened 2023-01-23
Don't count unimplemented errors as extra errors in stats

Amareis:
**Description:**

More clearly distinction between unimplemented yet features and wrong current type checking.

### DIFF
--- a/crates/stc_ts_type_checker/tests/tsc.rs
+++ b/crates/stc_ts_type_checker/tests/tsc.rs
@@ -776,9 +776,11 @@ fn do_test(file_name: &Path, spec: TestSpec, use_target: bool) -> Result<(), Std
         Err(err) => err,
     };
 
-    let extra_err_count = extra_errors.len();
-    stats.required_error += expected_errors.len();
-    stats.extra_error += extra_err_count - unimplemented_count;
+    let expected_count = expected_errors.len();
+    let extra_count = extra_errors.len();
+    let true_extra_count = extra_count - unimplemented_count;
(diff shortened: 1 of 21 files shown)


kdy1:
I think we should include them in the stat

kdy1:
Ah, I saw the changes

Amareis:
Add custom Debug impl for Stats, to not blow-up changes - unimplemented count is printed out only if there is any.

crates/stc_ts_type_checker/tests/tsc.rs
```diff
@@ -755,7 +778,8 @@ fn do_test(file_name: &Path, spec: TestSpec, use_target: bool) -> Result<(), Std
 
     let extra_err_count = extra_errors.len();
     stats.required_error += expected_errors.len();
-    stats.extra_error += extra_err_count;
+    stats.extra_error += extra_err_count - unimplemented_count;
+    stats.unimplemented = unimplemented_count;
```
Amareis, line 782: I don't really know if it should be handled more structurally - with removing such errors from extra errors vector and handling it differently in debug output below. I can do it, if it's needed.

kdy1 (changes requested):
I'm sorry, but I think this would interfere development

Amareis:
Fixed

kdy1:
I'm not going to accept this, and not sure what you mean by `fixed`

Amareis:
Oh, just thinks you sayed about merge conflict. Okay, close it now.

PR #556, opened 2023-01-24
fix(ts/analyzer): Handle TS2804

CarrotzRule123:
**Description:**

Handles duplicate static and instance names.

```typescript
class A {
   #b = 1
   static #b = 1 //TS2804
}
```


kdy1 (approved):
Thank you!

PR #567, opened 2023-01-28
feat: Handle assignment of a tuple to a tuple

kdy1:
**Description:**

 - Infer a type parameter as a single-element tuple if the parameter is of a rest type.
 - Use `access_property` for the assignment of tuples.
 - Use `access_property` for the inference of tuples.
 - Improve `access_property` for tuples.


kdy1:
This was wrong

Issue #552, opened 2023-01-24
Track destructuring


kdy1:
```ts
type Data = { kind: 'str', payload: string } | { kind: 'num', payload: number };

function gg2(obj: Data) {
    if (obj.kind === 'str') {
        let t: string = obj.payload;
    }
    else {
        let t: number = obj.payload;
    }
}

function foo({ kind, payload }: Data) {
    if (kind === 'str') {
        let t: string = payload;
    }
    else {
        let t: number = payload;
    }
}
```

Currently, the code above is handled properly, while the below is not. I have thought about this for a while and concluded we should handle this by storing a hashmap like `FxHashMap<Id, FakeName>` in the analyzer.

If you are willing to work on this, I'll try to expand this

---

Ideas:
 - `VarPath` for elements of the fake name.

---

I think looking at the source code of `tsc` would worth the effort.

sunrabbit123:
I will take it, But I can't guarantee it.

kdy1:
You can store an increment-only counter in the `Analyzer` and use it to distinguish patterns. One pattern (object/array) takes one id. Then, you can use it to *connect* identifiers.

PR #572, opened 2023-01-29
feat: Support a type guard with destructured vars

sunrabbit123:
**Description:**

```ts
//@strict: true

type Action = { kind: "A"; payload: number } | { kind: "B"; payload: string };

function f14<T extends Action>(t: T) {
  const { kind, payload } = t;
  if (kind === "A") {
    payload.toFixed();
  }
  if (kind === "B") {
    payload.toUpperCase();
  }
}
```

Previously, type guard for destructured allocation was not performed, so each variable was evaluated separately.

#552


crates/stc_ts_types/src/type_id/destructure.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use stc_visit::Visit;
+use swc_common::{EqIgnoreSpan, TypeEq};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Visit)]
+pub struct DestructureId(u32);
+
+impl DestructureId {
+    pub fn generate() -> Self {
+        static GENERATOR: AtomicU32 = AtomicU32::new(1);
```
kdy1, line 11: Do we really need atomics?

kdy1: I think this can be a waste

crates/stc_ts_types/src/lib.rs
```diff
@@ -281,7 +281,7 @@ impl Clone for Type {
 }
 
 #[cfg(target_pointer_width = "64")]
-assert_eq_size!(Type, [u8; 104]);
+assert_eq_size!(Type, [u8; 120]);
```
kdy1, line 284: Also, is this unevitable?

kdy1 (commented):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
```diff
@@ -1043,6 +1032,36 @@ impl Analyzer<'_, '_> {
         })
         .context("tried to ensure iterator")
     }
+
+    fn regist_destructure(&mut self, span: Span, ty: Option<&Type>) -> u32 {
+        match ty {
```
kdy1, line 1037: Please use `ty.as_ref().map(Type::normalize)` instead

kdy1 (changes requested):

crates/stc_ts_types/src/type_id/destructure.rs
sunrabbit123: I think the value should be unique even if the scope is different.
This is because the values may overlap.

crates/stc_ts_types/src/lib.rs
sunrabbit123: If you want to add a value to CommonMetadata, you must.
It contains metadata about the values before they were assigned unstructured.

crates/stc_ts_types/src/type_id/destructure.rs
kdy1: I mean, the analyzer is single threaded anyway, and we don't have to worry about conflicts between two or more modules

crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
sunrabbit123: Are you talking about this?
```rs
    fn regist_destructure(&mut self, span: Span, ty: Option<&Type>) -> u32 {
        match ty.as_ref().map(Type::normalize) {
```

kdy1: Yes

crates/stc_ts_types/src/type_id/destructure.rs
sunrabbit123: Come to think of it, I think it would be good to go back to the parent scope from the current scope to find the parent.

Postscript
There was a case of concern, but I can't remember.

kdy1: We can use a monotonicallly increasing numbe by using `Rc<Cell<u32>>`

sunrabbit123: Does your suggestion use the number of pointers in Rc as a counter?

crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
sunrabbit123, line 1037:
```suggestion
        match ty.as_ref().map(Type::normalize) {
```

crates/stc_ts_types/src/type_id/destructure.rs
kdy1: No, `Cell<u32>` is the value

sunrabbit123: Is this what you wanted?

```rs
impl DestructureId {
    thread_local! {
        static LAST_ID: Rc<Cell<u32>> = Rc::new(Cell::new(0));
    }

    pub fn generate() -> Self {
        let id = DestructureId::LAST_ID.with(|id| {
            let last_id = id.get().add(1);
            id.set(last_id);
            last_id
        });

        DestructureId(id)
    }

    pub fn get(id: u32) -> Self {
        DestructureId(id)
    }

    pub fn extract(&self) -> u32 {
        self.0
    }
}
```

kdy1: I meant storing a counter in the `Analyzer`

kdy1 (commented):
I'm not sure about the current way.

I think we can use an alternative type in https://github.com/dudykr/stc/blob/14d5ac24b3c4324d6384ae5d207f4fef9e3ae1fb/crates/stc_ts_file_analyzer/src/analyzer/control_flow.rs#L47-L49 . Instead of `Name`, we can change the key type, to prevent mistakes and to reduce the size of `Type`


But I'm not sure if it's better

sunrabbit123:
type guard with destructured vars support that top level is union.

Therefore, I think that the parent type should have Union and deal with it, not all types of Union.
destructured assigned with tuple is also going to be carried out with a label attached.

and

already developed `facts`, `vars`, `excludes` with `normal var`

As long as destructured vars is `var`, I think it's better to deal with it like this.

crates/stc_ts_types/src/type_id/destructure.rs
sunrabbit123: I will proceed after finishing the tuple.
thx

kdy1:
I see, then I think this is fine. I didn't know that only top-level destructurings were target

sunrabbit123:
I think it would be good to refer to this.

crates/stc_ts_types/src/type_id/destructure.rs
sunrabbit123: fixed it

crates/stc_ts_types/src/name.rs
```diff
@@ -20,6 +20,14 @@ type Inner = SmallVec<[Id; 4]>;
 pub struct Name(Inner);
 
 impl Name {
+    pub fn new(ctxt: SyntaxContext, name: JsWord) -> Self {
```
kdy1, line 23: Please change order

crates/stc_ts_types/src/type_id/destructure.rs
```diff
@@ -0,0 +1,27 @@
+use stc_visit::Visit;
+use swc_common::{EqIgnoreSpan, TypeEq};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Visit)]
+pub struct DestructureId(u32);
+
+impl DestructureId {
+    pub fn get(id: u32) -> Self {
```
kdy1, line 8: This function makes this type useless. Why do we need `extract`/`get` at first?

crates/stc_ts_types/src/type_id/destructure.rs
```diff
@@ -0,0 +1,27 @@
...
+        DestructureId(id)
+    }
+
+    pub fn extract(&self) -> u32 {
+        self.0
+    }
+}
+
+impl TypeEq for DestructureId {
```
kdy1, line 17: This should be `derive`-d instead

crates/stc_ts_types/src/type_id/destructure.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn type_eq(&self, other: &Self) -> bool {
+        self == other
+    }
+}
+
+impl EqIgnoreSpan for DestructureId {
```
kdy1, line 23: This should be `derive`-d instead

crates/stc_ts_types/src/lib.rs
```diff
@@ -1062,6 +1062,18 @@ impl TypeElement {
             TypeElement::Index(_) => None,
         }
     }
+
+    pub fn get_type(&self) -> Option<Type> {
```
kdy1, line 1066: Please change return type to `Option<&Type>`

kdy1 (changes requested):
Thanks!

crates/stc_ts_file_analyzer/src/analyzer/mod.rs
```diff
@@ -263,6 +265,8 @@ pub struct Analyzer<'scope, 'b> {
     debugger: Option<Debugger>,
 
     data: AnalyzerData,
+
+    destructure_count: Rc<Cell<u32>>,
```
kdy1, line 269: I think this can be `Rc<Cell<DestructureId>>`

crates/stc_ts_types/src/name.rs
sunrabbit123: Are there criteria in the order of factor values?

crates/stc_ts_file_analyzer/src/analyzer/mod.rs
```diff
@@ -537,6 +541,7 @@ impl<'scope, 'b> Analyzer<'scope, 'b> {
             imports_by_id: Default::default(),
             debugger,
             data,
+            destructure_count: Rc::new(Cell::new(0)),
```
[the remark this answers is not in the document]
sunrabbit123: want it?
```rs
Rc::new(Cell::new(Default::default()))
```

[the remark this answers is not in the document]
kdy1: No, I mean `Default::default()`.
`Rc` implements `Default` if the inner type implements `Default`

crates/stc_ts_types/src/lib.rs
sunrabbit123:
```rs
let ty: &Option<Box<Type>> = match self {
            TypeElement::Call(CallSignature { ret_ty, .. }) => ret_ty,
            TypeElement::Constructor(ConstructorSignature { ret_ty, .. }) => ret_ty,
            TypeElement::Property(PropertySignature { type_ann, .. }) => type_ann,
            TypeElement::Method(MethodSignature { ret_ty, .. }) => ret_ty,
            TypeElement::Index(IndexSignature { type_ann, .. }) => type_ann,
        };

        ty.as_ref().map(|ty: &Box<Type>| *ty.clone())
```
umm... Could you reconsider?

crates/stc_ts_types/src/metadata.rs
```diff
@@ -105,6 +105,8 @@ pub struct CommonTypeMetadata {
 
     /// TODO(kdy1): Move this to [TupleMetadata]
     pub prevent_tuple_to_array: bool,
+
+    pub destructure_key: u32,
```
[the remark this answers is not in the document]
sunrabbit123: [https://github.com/sunrabbit123/stc/blob/analyzer/type_guards_of_destructor/crates/stc_ts_file_analyzer/src/analyzer/scope/mod.rs#L1187-L1191](https://github.com/sunrabbit123/stc/blob/analyzer/type_guards_of_destructor/crates/stc_ts_file_analyzer/src/analyzer/scope/mod.rs#L1187-L1191)
is this?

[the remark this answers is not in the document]
kdy1: `pub destructure_key: DestructureId `

and I meant

```rust

impl DestructureId {

    fn next_id(&mut self) -> Self {
        self.0 += 1;
        *self
    }
}


```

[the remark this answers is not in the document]
kdy1: It's not changing about the size. It's about type safety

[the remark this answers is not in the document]
sunrabbit123: I applyed this
```rs
impl DestructureId {

    fn next_id(&mut self) -> Self {
        self.0 += 1;
        *self
    }
}
```
butI don't think I can handle the error after the change.  `pub destructure_key: DestructureId`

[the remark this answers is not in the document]
kdy1: Even if `next_id` is hard, you should `pub destructure_key: DestructureId` anyway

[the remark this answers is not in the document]
sunrabbit123: I'll check it out when I get home.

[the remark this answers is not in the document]
sunrabbit123: I meant not `next_id` is hard, `pub destructure_key: DestructureId` is hard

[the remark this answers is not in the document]
kdy1: Why? You declared the inner field using `pub`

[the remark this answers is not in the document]
sunrabbit123: I wll get this error
```
error[E0277]: the trait bound `type_id::destructure::DestructureId: _::_serde::Serialize` is not satisfied
    --> crates/stc_ts_types/src/metadata.rs:78:54
     |
78   | #[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
     |                                                      ^^^^^^^^^ the trait `_::_serde::Serialize` is not implemented for `type_id::destructure::DestructureId`
...
111  |     pub destructure_key: DestructureId,
     |     ---------------------------------- required by a bound introduced by this call
     |
     = help: the following other types implement trait `_::_serde::Serialize`:
               &'a T
               &'a mut T
               ()
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
               (T0, T1, T2, T3, T4)
               (T0, T1, T2, T3, T4, T5)
             and 685 others
note: required by a bound in `_::_serde::ser::SerializeStruct::serialize_field`
    --> /home/sunrabbit/.cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.152/src/ser/mod.rs:1901:12
     |
1901 |         T: Serialize;
     |            ^^^^^^^^^ required by this bound in `_::_serde::ser::SerializeStruct::serialize_field`
```

kdy1 (approved):
Thank you!

PR #582, opened 2023-01-31
feat: deep typeguard about destructor

sunrabbit123:
**Description:**
```ts
//@strict: true

type Func = <T extends ["a", number] | ["b", string]>(...args: T) => void;

const f60: Func = (kind, payload) => {
  if (kind === "a") {
    payload.toFixed(); // error
  }
  if (kind === "b") {
    payload.toUpperCase(); // error
  }
};
```

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/expr/function.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/expr/function.rs
@@ -217,9 +217,9 @@ impl Analyzer<'_, '_> {
             for (idx, param) in params.enumerate() {
                 if temp_els.len() == 1 {
                     if let Some(TupleElement { ty: box ty, .. }) = temp_els.first_mut() {
-                        if let Some(Type::Union(Union { types, .. })) = ty.as_union_type_mut() {
+                        if let Some(Union { types, .. }) = ty.as_union_type_mut() {
                             for ty in types {
-                                if let Some(Type::Tuple(Tuple { elems, .. })) = ty.as_tuple_mut() {
+                                if let Some(Tuple { elems, .. }) = ty.as_tuple_mut() {
                                     if idx < elems.len() {
                                         elems[idx].label = Some(param.clone());
                                     }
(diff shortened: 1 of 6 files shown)


crates/stc_ts_file_analyzer/src/analyzer/expr/function.rs
```diff
@@ -179,16 +179,56 @@ impl Analyzer<'_, '_> {
                         });
                     }
                 }
+                match (*param.ty).normalize() {
```
kdy1, line 182:
```suggestion
                match param.ty.normalize() {
```

crates/stc_ts_file_analyzer/src/analyzer/expr/function.rs
```diff
@@ -179,16 +179,56 @@ impl Analyzer<'_, '_> {
...
                 tracker: Default::default(),
             });
+            params_tuple.freeze();
+
+            let destructure_key = self.get_destructor_unique_key();
 
             for (idx, param) in params.enumerate() {
+                if temp_els.len() == 1 {
+                    if let Some(TupleElement { ty: box ty, .. }) = temp_els.first_mut() {
+                        if let Type::Union(Union { types, .. }) = ty.normalize_mut() {
```
kdy1, line 220: Please use `if let Some(Union { types, .. }) = ty.as_union_type_mut()` instead

crates/stc_ts_file_analyzer/src/analyzer/expr/function.rs
```diff
@@ -179,16 +179,56 @@ impl Analyzer<'_, '_> {
...
+                            for ty in types {
+                                if let Type::Tuple(Tuple { elems, .. }) = ty.normalize_mut() {
```
kdy1, line 222: Please use if let Some(Tuple { elems, .. }) = ty.as_union_type_mut() instead

kdy1 (changes requested):
Thank you!

sunrabbit123:
fixed it

kdy1 (dismissed):
Thanks!

kdy1 (approved):

PR #608, opened 2023-02-03
Misc(.gitignore/analyzer): ignore typescript from scripts/check.sh

sunrabbit123:
**Description:**

If you run it for some reason, the file will be created.
But it's so big and beautiful that I can't put my git in it.
Ignore it because it is not a useful file.


sunrabbit123:
sorry I used `script` bash command
I fault

PR #609, opened 2023-02-03
feat: Improve handling of deep type guards

sunrabbit123:
**Description:**

**BREAKING CHANGE:**

The following forms have been processed.
```ts
//@strict: true

type ReducerArgs =
  | ["add", { a: number; b: number }]
  | ["concat", { firstArr: any[]; secondArr: any[] }];

const reducer: (...args: ReducerArgs) => void = (op, args) => {
  switch (op) {
    case "add":
      console.log(args.a + args.b);
      break;
    case "concat":
      console.log(args.firstArr.concat(args.secondArr));
      break;
  }
};
```


kdy1 (approved):
Thank you!

PR #621, opened 2023-02-04
fix: Fix false positives related to generators

kdy1:
**Description:**


kdy1:
Postponed due to https://github.com/dudykr/stc/pull/624

PR #653, opened 2023-02-09
update: enum error accuracy

sunrabbit123:
I will work next situation

```ts
enum T1 {
    a = "1",
    e = "a" + 1  // error
}
```


crates/stc_ts_file_analyzer/src/analyzer/enums.rs
```diff
@@ -655,3 +671,14 @@ impl Visit<RExpr> for LitValidator<'_> {
         }
     }
 }
+
+fn has_not_str_and_not_plus(bin: &RBinExpr) -> bool {
```
kdy1, line 675: Can you rename this to `non_str_nor_plus`?

kdy1 (changes requested):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/enums.rs
sunrabbit123: fixed it

kdy1 (approved):
Thanks!

PR #656, opened 2023-02-10
Feat: check ImplicitAnyBecauseIndexTypeIsWrong errors

sunrabbit123:
**Description:**

1. I add some option about base test
2. I increase extra_error on objectSpreadIndexSignature
    Because readonly flag not set false on object spread

thx.


kdy1 (approved):
Thank you!

I inverted the PR to fix the regression.

Issue #594, opened 2023-02-02
Index signature with string indexer should be assignable to one with numeric indexer


kdy1:
https://github.com/dudykr/stc/blob/273356f921c6152b5a45c41527923a86361ebe13/crates/stc_ts_type_checker/tests/conformance/types/typeRelationships/typeInference/indexSignatureTypeInference.ts#L1-L17

At the code above, line 17 should pass. i.e. `StringMap<Function>` is assignable to `NumberMap<Function>`.

YoniFeng:
issue can be closed

kdy1:
Thank you!

PR #660, opened 2023-02-10
fix: Index signature type inference

Cozzub:
**Description:**
 Allow infering types from string map interface arg to number map param https://github.com/dudykr/stc/issues/594


Cozzub:
I`m not sure when and how create test files in crates/stc_ts_file_analyzer/tests/pass | pass-only.  Are there any explanation about it?  Should i add my case also?

sunrabbit123:
Creating file is not required.

If you are going to create a file, you should place it similar to the path to the original file.
In my case, I tend to remove and create duplicate names.

kdy1:
Is this PR ready for a review?

kdy1 (approved):
Thank you!

Issue #592, opened 2023-02-01
Change error code for non-iterable based on the target


kdy1:
https://github.com/dudykr/stc/blob/8b5fad3b580e2a4e2a4a75b0e3f45b75a8cc3f63/crates/stc_ts_type_checker/tests/conformance/es6/destructuring/destructuringAssignabilityCheck.ts#L5

We should report [`TS2461`](https://github.com/microsoft/TypeScript/blob/26eced32ed464c2d31cd6007918224ec0df64c18/src/compiler/diagnosticMessages.json#L2259-L2262) instead of `TS2345` for the line above

awareness481:
I think an issue is that https://github.com/dudykr/stc/blob/8ea5b02590550cdd7d1e6cbf53a54bc719be99a2/crates/stc_ts_file_analyzer/src/analyzer/expr/array.rs#L605 is never called, which by running the conformance tests on the `microsoft/TypeScript` repo I figured should be called for this to be handled correctly. I couldn't figure out why but I suspect this might be relevant:

I modified the `fail` macro to log `to` and `from` and got back (for `to`)
```rs
Tuple(Tuple { span: Span { lo: BytePos(3), hi: BytePos(5), ctxt: #0 }, elems: [], metadata: TupleMetadata { prevent_tuple_to_array: false, common: CommonTypeMetadata { implicit:
  │ false, infected_by_this_in_object_literal: false, prevent_converting_to_children: false, contains_infer_type: false, prevent_complex_simplification: false, resolved_from_var: false,
  │ prevent_generalization: false, destructure_key: DestructureId(1) } }, tracker: Tracker { _priv: () } })
```

which to me is weird because I expected an array type but I dont understand `stc` well enough to know for sure. If it should have been an `array` type then that explains why `get_iterator_inner` is never called. 

I also had the following logged:

```rs
[crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs:93] &ty = Arc(
    Freezed {
        ty: Function(
            Function {
                span: Span {
                    lo: BytePos(
                        2,
                    ),
                    hi: BytePos(
                        11,
                    ),
                    ctxt: #0,
                },
                type_params: None,
                params: [
                    FnParam {
                        span: Span {
                            lo: BytePos(
                                3,
                            ),
                            hi: BytePos(
                                5,
                            ),
                            ctxt: #0,
                        },
                        required: true,
                        pat: Array(
                            RArrayPat {
                                node_id: NodeId(
                                    6,
[...]

PR #661, opened 2023-02-10
potential fix to #592

awareness481:
**Description:**

Potentially solves #592 . As far as I can tell, tuple types should be widened on iife if there's no type_ann . I tried to understand if there are other cases where they should be certainly widened but I couldn't find anything conclusive. Not sure if the changes are good enough to be mergeable. First time Im trying to contribute so I hope I didnt do anything horribly wrong.

Edit: I'm not sure if it's problematic that this error does not come from `get_iterator_inner` like in `tsc` but I couldn't figure out a way to do it from there

https://github.com/dudykr/stc/issues/592

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
@@ -3200,6 +3200,12 @@ impl Analyzer<'_, '_> {
                                 _ => {}
                             }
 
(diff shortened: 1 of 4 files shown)


kdy1 (approved):
Thank you!

kdy1:
> Edit: I'm not sure if it's problematic that this error does not come from get_iterator_inner like in tsc but I couldn't figure out a way to do it from there

I'm not sure at the moment, too. It should be investigated

PR #673, opened 2023-02-11
Feat: implement a bit of priority about index signature

sunrabbit123:
**Description:**
```ts
const sym = Symbol()
type Both = { s: number, '0': number, [sym]: boolean } | { [n: number]: number, 1: 1; [s: string]: string | number }
declare var both: Both
both[0] = 'not ok'
```


kdy1 (approved):
Thank you!

PR #674, opened 2023-02-11
partially implements logic for conformance/types/unknown/unknownType1.ts

awareness481:
**Description:**

[conformance/types/unknown/unknownType1.ts](https://github.com/dudykr/stc/blob/main/crates/stc_ts_type_checker/tests/conformance/types/unknown/unknownType1.ts)

`f30`, `f23`, `f24` and `class C1` are still not implemented


crates/stc_ts_file_analyzer/src/analyzer/expr/object.rs
```diff
@@ -39,6 +39,16 @@ impl Analyzer<'_, '_> {
             let mut known_keys = vec![];
             for prop in node.props.iter() {
                 ret = a.append_prop_or_spread_to_type(&mut known_keys, ret, prop, type_ann.as_deref())?;
+
+                if ret.is_unknown() {
+                    if let RPropOrSpread::Spread(..) = prop {
+                        return Err(ErrorKind::NonObjectInSpread {
```
awareness481, line 45: I think this needs to be `self.storage.report` because otherwise it won't catch 2 unknown spread elements ex:

```ts
function fn1(a: unknown, b: unknown) {
  let o1 = { ...a, ...b };
}
```

currently only `a` gets the error but TypeScript (from the IDE) seems to add the error to both `a` and `b`

crates/stc_ts_errors/src/lib.rs
```diff
@@ -2069,6 +2078,10 @@ impl ErrorKind {
 
             ErrorKind::ExportAmbientModule { .. } => 2668,
 
+            ErrorKind::IsTypeUnknown { .. } => 18046,
```
awareness481, line 2081: Not sure about the error names here

crates/stc_ts_file_analyzer/src/analyzer/expr/object.rs
kdy1: I think you are right

crates/stc_ts_errors/src/lib.rs
kdy1: It's fine I think

crates/stc_ts_file_analyzer/src/analyzer/expr/object.rs
awareness481: I tried with `self.storage.report` and `self.storage.report_all` but I still got the same results. As in, neither of them would report the error on `...d` here, only on the first one

```ts
function f26(x: {}, y: unknown, z: any, d: unknown) {
  let o2 = { a: 42, ...y, ...d, ...x }; // unknown
  // only ...y is reported by stc
}
```

awareness481:
@kdy1 I spent more than 5 hours yesterday trying to solve [f30](https://github.com/dudykr/stc/blob/b5f1cf11908d89b2a82854249e6f296a0cb5109a/crates/stc_ts_type_checker/tests/conformance/types/unknown/unknownType1.ts#L171), [f23](https://github.com/dudykr/stc/blob/b5f1cf11908d89b2a82854249e6f296a0cb5109a/crates/stc_ts_type_checker/tests/conformance/types/unknown/unknownType1.ts#L121), [ f24 ](https://github.com/dudykr/stc/blob/b5f1cf11908d89b2a82854249e6f296a0cb5109a/crates/stc_ts_type_checker/tests/conformance/types/unknown/unknownType1.ts#L127) but I would only find hacky solutions for them. I think they're all the same case and they probably need to be handled via `self.extends` but I don't know how the architecture for that would look like.

I think I will be able to solve [C1](https://github.com/dudykr/stc/blob/b5f1cf11908d89b2a82854249e6f296a0cb5109a/crates/stc_ts_type_checker/tests/conformance/types/unknown/unknownType1.ts#L163) in the future, but I'm spent for the weekend, so I propose we can move with the PR as is. 

There's also [this](https://github.com/dudykr/stc/pull/674#discussion_r1103765654)

crates/stc_ts_file_analyzer/src/analyzer/expr/object.rs
kdy1: My rough guess is that you used the same span, but I think it's fine to postpone it for another PR.

kdy1 (approved):
Thank you!

PR #675, opened 2023-02-11
feat: Instantiate return type while validating

kdy1:
**Description:**

 - Instantiate types while validating return types.
 - Improve `get_awaited_type`.
 - Optimize assignment of `Promise`.


crates/stc_ts_type_checker/tests/conformance/types/specifyingTypes/typeQueries/circularTypeofWithVarOrFunc.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
     required_error: 10,
     matched_error: 0,
-    extra_error: 2,
+    extra_error: 3,
```
kdy1, line 4: Investigation:

This is caused by the cycle not being detected.

kdy1: https://github.com/dudykr/stc/issues/361

PR #683, opened 2023-02-12
attempt to fix error reports on non-object spread elements

awareness481:
**Description:**

[More context](https://github.com/dudykr/stc/pull/674#discussion_r1103621630)

prev:
```ts
function f26(x: {}, y: unknown, z: any, d: unknown) {
  let o2 = { a: 42, ...y, ...d, ...x }; // unknown
                      __
  // only y gets an error
}
```
after:
```ts
function f26(x: {}, y: unknown, z: any, d: unknown) {
  let o2 = { a: 42, ...y, ...d, ...x }; // unknown
                     __   __
  // both y and d get the error
}
```

I also noticed that using `ret.is_unknown()` was probably problematic. If we changed the order of parameters like this:

```ts
// move x to last parameter
function f26(y: unknown, z: any, d: unknown, x: {}) {
  let o2 = { a: 42, ...y, ...d, ...x }; // unknown
}
```
`ret.is_unknown()` would be `true` for `x`. Which is why I ended up using `self.type_of_raw_var`. I didn't know if I should open an issue or open a PR so please feel free to reject the PR is this solution isn't good enough

**Edit**: Now that I think about it, `self.type_of_raw_var` gets called once already (based on the debug logs) so this solution is probably a bit wasteful (makes it get called twice)


kdy1:
Do you mean the type of the object is already `unknown` when adding `x`?

crates/stc_ts_file_analyzer/src/analyzer/expr/object.rs
```diff
@@ -40,13 +40,26 @@ impl Analyzer<'_, '_> {
             for prop in node.props.iter() {
                 ret = a.append_prop_or_spread_to_type(&mut known_keys, ret, prop, type_ann.as_deref())?;
 
-                if ret.is_unknown() {
-                    if let RPropOrSpread::Spread(..) = prop {
-                        return Err(ErrorKind::NonObjectInSpread {
-                            span: ret.span(),
-                            ty: Box::new(ret),
+                if let RPropOrSpread::Spread(element) = prop {
+                    if let RExpr::Ident(some) = &*element.expr {
```
kdy1, line 44: I think we should not match the AST type of the expression.

kdy1 (changes requested):
Thanks!
Calling `type_of_raw_var` twice is fine, but I think this is problematic because this will not work for other kinds of expressions, like member or call.

awareness481:
> Do you mean the type of the object is already `unknown` when adding `x`?

If you replace the `for loop` in the PR with this:

``rs
            for prop in node.props.iter() {
                ret = a.append_prop_or_spread_to_type(&mut known_keys, ret, prop, type_ann.as_deref())?;

                if ret.is_unknown() {
                    if let RPropOrSpread::Spread(element) = prop {
                        a.storage.report(
                            ErrorKind::NonObjectInSpread {
                                span: Span {
                                    lo: element.dot3_token.span().lo,
                                    hi: element.span().hi,
                                    ctxt: SyntaxContext::empty(),
                                },
                                ty: Box::new(Type::Keyword(KeywordType {
                                    kind: TsKeywordTypeKind::TsUnknownKeyword,
                                    span: element.span(),
                                    metadata: Default::default(),
                                    tracker: Default::default(),
                                })),
                            }
                            .into(),
                        );
                    }
                }
            }
```

and have as input
```ts
function f26(x: {}, y: unknown, z: any, d: unknown) {
  let o2 = { a: 42, ...y, ...d, ...x }; // unknown
}
```
you will get

```rs
  × NonObjectInSpread {
  │     span: Span {
  │         lo: BytePos(
  │             3756,
  │         ),
  │         hi: BytePos(
  │             3760,
  │         ),
  │         ctxt: #0,
  │     },
  │     ty: unknown,
  │ }
     ╭─[./a.ts:143:1]
 143 │ function f26(x: {}, y: unknown, z: any, d: unknown) {
 144 │   let o2 = { a: 42, ...y, ...d, ...x }; // unknown
     ·                                 ────
 145 │ }
```

which is why I tried something different :P .

kdy1:
Can't we return two values from `append_prop_or_spread_to_type`?
The first return value is type of object, and the second return value is `Option<Type>` which is `Some` if `prop` is a spread.

kdy1:
Or alternatively we can report the errors from `append_prop_or_spread_to_type`

awareness481:
> Or alternatively we can report the errors from `append_prop_or_spread_to_type`

Ohhh I think we can, I had already tried that but I was using the `to` variable. It should work with `prop_ty`. Let me get back to you in a bit

kdy1:
Seems like there's no change in stats?

kdy1:
Can you add some test cases in [`tsc`](https://stc.dudy.dev/docs/contributing#teststsc)?

awareness481:
> Seems like there's no change in stats?

There shouldn't be (at least for `unknownType1.ts`). The conformance test case only covers one unknown 

```ts
function f26(x: {}, y: unknown, z: any) {
    let o1 = { a: 42, ...x };  // { a: number }
    let o2 = { a: 42, ...x, ...y };  // unknown
    let o3 = { a: 42, ...x, ...y, ...z };  // any
    let o4 = { a: 42, ...z }; // any
}
```

The issue with the 2 unknowns was something I discovered while trying to make `report_all` work

> Can you add some test cases in [`tsc`](https://stc.dudy.dev/docs/contributing#teststsc)?

Yes sure, but that'll be in at least a couple days.

kdy1:
No worries, I can add them. :)
I'll add them and merge PR

kdy1 (approved):
It worked!
Thank you!

PR #687, opened 2023-02-12
fixed type checking when class extends class with private fields

infix:
**Description:**

we previously only handled this case for interfaces, now we added the check to classes


crates/stc_ts_errors/src/lib.rs
```diff
@@ -1788,6 +1788,8 @@ impl ErrorKind {
             ErrorKind::TypeInvalidForUpdateArg { .. } => 2356,
             ErrorKind::ExprInvalidForUpdateArg { .. } => 2357,
 
+            ErrorKind::PrivatePropertyIsDifferent { .. } => 2415,
```
infix, line 1791: i thought about creating a new ErrorKind here but it made sense to reuse it. it previously was used for emit 2420 for incorrectly implementing interfaces. so this might be wrong

crates/stc_ts_file_analyzer/src/analyzer/class/mod.rs
```diff
@@ -1442,6 +1442,13 @@ impl Analyzer<'_, '_> {
                                             .report(ErrorKind::DefinedWithAccessorInSuper { span: p.key.span() }.into())
                                     }
 
+                                    if super_property.accessibility == Some(Accessibility::Private)
+                                        && p.accessibility != Some(Accessibility::Private)
+                                    {
+                                        self.storage
+                                            .report(ErrorKind::PrivatePropertyIsDifferent { span: super_ty.span() }.into())
```
infix, line 1449: not sure if `super_ty.span()` is the best way to get the parent classes name

kdy1: It's the best way

kdy1 (approved):
Thank you!
I inverted the PR to make it pass CI

infix:
Awesome, thank you! I was actually confused regarding why the CI failing :)

PR #688, opened 2023-02-12
fix: Fix normalization of an indexed access type with type parameter key

kdy1:
**Description:**

### DIFF
--- a/crates/stc_ts_type_checker/tests/conformance/types/conditional/conditionalTypes2.stats.rust-debug
+++ b/crates/stc_ts_type_checker/tests/conformance/types/conditional/conditionalTypes2.stats.rust-debug
@@ -1,6 +1,6 @@
 Stats {
     required_error: 2,
     matched_error: 5,
-    extra_error: 5,
+    extra_error: 6,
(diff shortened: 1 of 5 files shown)


crates/stc_ts_type_checker/tests/conformance/types/conditional/conditionalTypes2.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
     required_error: 2,
     matched_error: 5,
-    extra_error: 5,
+    extra_error: 6,
```
kdy1, line 4: This regression is due to previous behavior

kdy1: We returned `any` so errors were not validated correctly

kdy1: I extracted this PR from https://github.com/dudykr/stc/pull/686 because of this regression

PR #689, opened 2023-02-13
Fix: right hand side of statement must be any

sunrabbit123:
**Description:**

fix next
```ts
declare function f4(): never;

for (const n in f4()) {}
```

But, I increase 1 error
I don't understand this error

follow the next code
```ts
for (var x in 42 ? d[x] : c[x]) { }
```
It's not written in error.json.
There should be an error. I think.

### DIFF
--- a/crates/stc_ts_type_checker/tests/conformance/statements/for-inStatements/for-inStatementsInvalid.stats.rust-debug
+++ b/crates/stc_ts_type_checker/tests/conformance/statements/for-inStatements/for-inStatementsInvalid.stats.rust-debug
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 14,
-    matched_error: 3,
-    extra_error: 1,
+    required_error: 6,
+    matched_error: 11,
+    extra_error: 2,
     panic: 0,
 }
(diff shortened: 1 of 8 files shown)


kdy1 (approved):
Thank you!
I'll investigate the regression

crates/stc_ts_type_checker/tests/conformance/statements/for-inStatements/for-inStatementsInvalid.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 14,
-    matched_error: 3,
-    extra_error: 1,
+    required_error: 6,
+    matched_error: 11,
+    extra_error: 2,
```
kdy1, line 4: This is caused by the multiple `x`.

kdy1: image
image

kdy1: I think this is crazy... I'll merge it.

kdy1 (approved):

kdy1 (approved):

PR #690, opened 2023-02-13
Fix: A function returning 'never' cannot have a reachable end point.

sunrabbit123:
**Description:**

Fix next code
```ts
function f4() : never {
}
```

But, increase 2 extra error on next code
```ts
class MyThrowable {
    throw(): never {
        throw new Error();
    }
}

class SuperThrowable extends MyThrowable {
    err(msg: string): never {
        super.throw() // this
    }
    ok(): never {
        this.throw() // this
    }
}
```

because we can't have check this when check function return type with stmt.
Do you have any good suggestions?


crates/stc_ts_file_analyzer/src/analyzer/stmt/return_type.rs
```diff
@@ -72,10 +74,34 @@ impl Analyzer<'_, '_> {
...
         for stmt in stmts {
-            if let RStmt::Throw(throws) = stmt {
-                unconditional_throw = Some(throws.span);
-                break;
+            match stmt {
+                RStmt::Throw(throws) => {
+                    unconditional_throw = Some(throws.span);
+                    break;
+                }
+                RStmt::While(RWhileStmt {
```
kdy1, line 84: Is this logic inevitable?

kdy1: I think handling AST from here is a too bad for future maintenance

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/stmt/return_type.rs
sunrabbit123: If so, is there a device that detects the code that causes never within the block in the process of obtaining the return type?

sunrabbit123: If possible, it will also be possible to erase the added extra_error.

kdy1: I think we should add a field to `ReturnValues` and modify the value from validator for children nodes

sunrabbit123: I think it's no different Both my solution and fn check_for_infinite_loop

Even in_reachable is not spreading from fn check_for_infinite_loop

sunrabbit123: It's finally spreading.
By the way, why are repeat statements also checked in_conditional?

sunrabbit123: https://github.com/dudykr/stc/blob/main/crates/stc_ts_file_analyzer/src/analyzer/stmt/mod.rs#L54

kdy1: > By the way, why are loop statements also checked in_conditional?

I think it's my mistake

kdy1: Alternatively, we can add field to the `Ctx` like `in_never_returning_function` and use it from the validator or an expression for a statement

sunrabbit123: I think that better solution is spread is_unreachagble
how do you think?

kdy1: I'm not sure what is spreading `in_unreachable`

sunrabbit123: umm.... propagate ?

kdy1: I think it can be better 👍

sunrabbit123: I fixed it

kdy1: Thank you!

crates/stc_ts_file_analyzer/src/analyzer/stmt/loops.rs
```diff
@@ -89,7 +90,16 @@ impl Analyzer<'_, '_> {
...
         self.cur_facts.true_facts += prev_facts;
         self.cur_facts.false_facts += prev_false_facts;
-
+        if !self.scope.return_values.in_conditional {
+            let mut v = LoopBreakerFinder { found: false };
+            body.visit_with(&mut v);
+            let has_break = v.found;
+            if !has_break {
+                if let Some(RExpr::Lit(RLit::Bool(RBool { value, .. }))) = test {
+                    self.ctx.in_unreachable = *value;
```
kdy1, line 99: Isn't this `!*value`?

`while (true)` => reachable
`while (false)` => unreachable

sunrabbit123: `while true` is unreachable

sunrabbit123: if infinite loop in scope, we can get unreachable
infiinite loop comes from `condition statement` is always `true`

kdy1: Ah, I see. Thanks!

kdy1 (approved):
Thank you!

PR #692, opened 2023-02-13
[conformance/expressions/functionCalls/callWithMissingVoid.ts] - correctly count void in params

awareness481:
**Description:**

Correctly handle `void` when in parameters. Don't add it in `min_params` count if it's the last one or all next params are `void`. Make sure to handle `void -> unknown` when `unknown` is param.

Only **not fixed** case is:
```ts
declare function call<TS extends unknown[]>(
    handler: (...args: TS) => unknown,
    ...args: TS): void;

call((x: number, y: number) => x + y) // error
```
but I dont think I have the time to fix it in this PR


crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
```diff
@@ -2252,33 +2262,6 @@ impl Analyzer<'_, '_> {
                 }
                 _ => {}
             }
-            if param.required {
```
awareness481, line 2255: If I understand correctly with my change we don't need this, and removing this piece of code makes more tests pass so ¯\_(ツ)_/¯

awareness481: Ended up removing this in favor of handling `void` params at the beginning.

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -1122,7 +1122,22 @@ impl Analyzer<'_, '_> {
             Type::Keyword(KeywordType {
                 kind: TsKeywordTypeKind::TsUnknownKeyword,
                 ..
-            }) => return Ok(()),
+            }) => {
+                // We need to do this because everything is assignable to unknown.
+                // But if we are in argument, we should not allow void (which means no argument
+                // was passed)
+                if self.ctx.in_argument {
```
kdy1, line 1129: Can we add a field to `AssignOpts` instead of using context?

kdy1:
```ts
call(() => {
    var v: void;
    const foo: string = v;
})
```

kdy1: This code is problematic if we use `self.ctx.in_argument`.
We can add a field to `AssignOpts`, e.g. `in_argument: bool` and set it to true while assigning for arguments.

crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
```diff
@@ -2211,9 +2211,25 @@ impl Analyzer<'_, '_> {
 
         let span = span.with_ctxt(SyntaxContext::empty());
 
-        let mut min_param: usize = params.iter().map(|v| &v.pat).map(count_required_pat).sum();
+        let min_param: usize = params
+            .iter()
+            .enumerate()
+            .filter_map(|(i, v)| {
+                if v.ty.contains_void() {
+                    let next_params = params[i + 1..params.len()].to_vec();
```
kdy1, line 2219: Is `to_vec()` necessary? It allocates, so I'm curious if `&params[i + 1..params.len()]` will work

kdy1 (changes requested):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: I used `allow_assignment_to_void: false` instead, I think that makes sense. I'm not sure why the correct error is reported without me doing anything else tho :thinking:

kdy1 (approved):
Thank you!

Issue #591, opened 2023-02-01
Change error code for mixed default exports


kdy1:
https://github.com/dudykr/stc/blob/8b5fad3b580e2a4e2a4a75b0e3f45b75a8cc3f63/crates/stc_ts_type_checker/tests/conformance/es6/modules/defaultExportsCannotMerge04.ts#L1-L15

We should report [`TS2652`](https://github.com/microsoft/TypeScript/blob/26eced32ed464c2d31cd6007918224ec0df64c18/src/compiler/diagnosticMessages.json#L2896) instead of `TS2395`

filipw01:
@kdy1 Can I try this one?

kdy1:
Of course, thank you!

PR #695, opened 2023-02-14
fix: Fix error code for mixed default exports

filipw01:
**Description:**
Adds TS2652

I have to admit my approach is kind of naive and only fixes the `defaultExportsCannotMerge04` so it will need to be edited to satisfy other `defaultExportsCannotMerge...` cases (for example `export default class`)

I can work on the other ones too, but
1. I don't know how to enable them because running `./scripts/test.sh defaultExportsCannotMerge` runs only `defaultExportsCannotMerge04`
2. I want to get some feedback as this is my first contribution in Rust

Should we make this one pretty and merge or should we tackle all `defaultExportsCannotMerge...` in one PR?

Closes: #591

### DIFF
--- a/crates/stc_ts_errors/src/lib.rs
+++ b/crates/stc_ts_errors/src/lib.rs
@@ -594,6 +594,12 @@ pub enum ErrorKind {
     ExportMixedWithLocal {
         span: Span,
     },
+
+    /// TS2652
+    MixedDefaultExports {
+        span: Span,
+    },
+
     /// TS2420
     ClassIncorrectlyImplementsInterface {
         span: Span,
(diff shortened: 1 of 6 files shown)


sunrabbit123:
1. It's the same for me when I run it.
	I'll create a PR that includes an alternative.
3. PR is OK to be separated.

sunrabbit123:
If @filename is included, we have confirmed that it will not be tested.

kdy1 (approved):
Thank you!

kdy1:
I think we can use another PR for further improvements this PR is good enough for merging.

kdy1:
I'm not sure why, but I can't enable auto-merge at the moment.

PR #696, opened 2023-02-15
fix: Fix the error code for a usage of property before init

togami2864:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #697, opened 2023-02-15
Fix: wrong assignment with compatting interface to function

sunrabbit123:
**Description:**

Fix next 
```ts
interface T {
    f(x: number): void;
}
var t: T;
t = () => 1;
t = function (x: number) { return ''; }
```


kdy1 (approved):
Thank you!

PR #727, opened 2023-02-18
fix: type predicates getting expanded to js_word and then Interface

awareness481:
**Description:**

The issue here is that `Object` and `Function` in type predicates were normalized to js_word and then converted to type `Interface`

```ts
declare function isObject(x: any): x is Object // -> x converted to Interface
declare function isFunction(x: any): x is Function // -> converted to Interface
```

That's why I added a `self.ctx.is_type_predicate`. I'm still not confident that my change is good enough or correct, but the following code seems to have no issues

```ts
declare function isTypeofObject(x: any): x is typeof Object;
declare function isUnionFunction(x: any): x is Function | number;
```


crates/stc_ts_file_analyzer/src/analyzer/scope/mod.rs
```diff
@@ -1151,8 +1151,10 @@ impl Analyzer<'_, '_> {
             }
         }
 
-        if let Ok(ty) = self.env.get_global_type(DUMMY_SP, name.sym()) {
-            return Ok(Some(ItemRef::Owned(vec![ty].into_iter())));
+        if !self.ctx.is_type_predicate {
+            if let Ok(ty) = self.env.get_global_type(DUMMY_SP, name.sym()) {
```
awareness481, line 1155: As far as I can tell this is the function that converts the type predicates to interfaces so thats why I added the if check here

crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
```diff
@@ -3290,9 +3293,27 @@ impl Analyzer<'_, '_> {
...
+                    Type::Keyword(KeywordType {
+                        kind: TsKeywordTypeKind::TsAnyKeyword,
+                        ..
+                    }) => {
+                        if new_ty.is_fn_type() {
+                            return Ok(orig_ty.into_owned());
+                        }
+
+                        return Ok(new_ty.into_owned());
+                    }
```
awareness481, line 3316: This logic is based on this comment [// 'any' is narrowed to types other than 'Function'/'Object' (including {})](https://github.com/microsoft/TypeScript/blob/d87d0adcd30ac285393bf3bfbbb4d94d50c4f3c9/tests/baselines/reference/narrowFromAnyWithTypePredicate.errors.txt#L29)

awareness481, line 3316: This logic is based on this comment [// 'any' is narrowed to types other than 'Function'/'Object' (including {})](https://github.com/microsoft/TypeScript/blob/d87d0adcd30ac285393bf3bfbbb4d94d50c4f3c9/tests/baselines/reference/narrowFromAnyWithTypePredicate.errors.txt#L29)

crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
```diff
@@ -3227,6 +3227,8 @@ impl Analyzer<'_, '_> {
         }
 
         if let Type::Predicate(p) = ret_ty.normalize() {
+            self.ctx.is_type_predicate = true;
```
kdy1, line 3230: Can we use `with_ctx` instead?

kdy1 (changes requested):
Thank you! I think we need to reset the `ctx`.

crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs
awareness481: That makes `pass_only_tests__pass_only__conformance__types__conditional__conditionalTypes2__3_ts
` fail with

(it fails on a function with a type predicate)

```rs
context: tried to validate an argument (at crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs:3206:25)
  │ context:
  │ lhs = instanceof  (object | Function);
  │ rhs = T#14#0 extends Function#0 ? T : never;
  │ lhs (normalized) = (object | Function#0);
  │ rhs (normalized) = T#14#0 extends Function#0 ? T : never; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:587:73)
  │ context: tried to assign the true type of a conditional type to lhs (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:1089:22)
  │ context:
  │ lhs = (object | Function#0);
  │ rhs = T#14#0;
  │ lhs (normalized) = (object | Function#0);
  │ rhs (normalized) = T#14#0; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:587:73)
  │ context: tried to assign a type to a union type (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:1985:67)
  │ WrongArgType {
  │     span: Span {
  │         lo: BytePos(
  │             1089,
  │         ),
  │         hi: BytePos(
  │             1094,
  │         ),
  │         ctxt: #14,
  │     },
  │     inner: Errors {
  │         span: Span {
  │             lo: BytePos(
  │                 1089,
  │             ),
  │             hi: BytePos(
  │                 1094,
  │             ),
  │             ctxt: #14,
  │         },
  │         errors: [
  │             context: tried to assign a type to a union (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:1951:26)
  │             context:
  │             lhs = object;
  │             rhs = T#14#0;
  │             lhs (normalized) = object;
  │             rhs (normalized) = T#14#0; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:587:73)
  │             context: LHS (final): object;
  │             RHS (final): T#14#0; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:1694:25)
  │             AssignFailed {
  │                 span: Span {
  │                     lo: BytePos(
  │                         1089,
  │                     ),
  │                     hi: BytePos(
  │                         1094,
  │                     ),
  │                     ctxt: #14,
  │                 },
  │                 cause: [],
  │             },
  │             context: tried to assign a type to a union (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:1951:26)
  │             context:
  │             lhs = Function#0;
  │             Member as {
  │                 apply(this: Function, thisArg: any, argArray?: any): any;
  │                 call(this: Function, thisArg: any, ...argArray: any[]): any;
  │                 bind(this: Function, thisArg: any, ...argArray: any[]): any;
  │                 toString(): string;
  │                 prototype: any;
  │                 readonly length: number;
  │                 arguments: any;
  │                 caller: Function;
  │             };
  │             rhs = T#14#0;
  │             lhs (normalized) = Function#0;
  │             Member as {
  │                 apply(this: Function, thisArg: any, argArray?: any): any;
  │                 call(this: Function, thisArg: any, ...argArray: any[]): any;
  │                 bind(this: Function, thisArg: any, ...argArray: any[]): any;
  │                 toString(): string;
  │                 prototype: any;
  │                 readonly length: number;
  │                 arguments: any;
  │                 caller: Function;
  │             };
  │             rhs (normalized) = T#14#0; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:587:73)
  │             context: LHS (final): Function#0;
  │             Member as {
  │                 apply(this: Function, thisArg: any, argArray?: any): any;
  │                 call(this: Function, thisArg: any, ...argArray: any[]): any;
  │                 bind(this: Function, thisArg: any, ...argArray: any[]): any;
  │                 toString(): string;
  │                 prototype: any;
  │                 readonly length: number;
  │                 arguments: any;
  │                 caller: Function;
  │             };
  │             RHS (final): T#14#0; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:1694:25)
  │             AssignFailed {
  │                 span: Span {
  │                     lo: BytePos(
  │                         1089,
  │                     ),
  │                     hi: BytePos(
  │                         1094,
  │                     ),
  │                     ctxt: #14,
  │                 },
  │                 cause: [],
  │             },
  │         ],
  │     },
  │ }
    ╭─[./a.ts:46:1]
 46 │     if (isFunction(value)) {
 47 │         toString1(value);
    ·                   ─────
 48 │         toString2(value);
    ╰────
TS2345

  × context: tried to validate an argument (at crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs:3206:25)
  │ context:
  │ lhs = instanceof  Function;
  │ rhs = T#14#0 extends Function#0 ? T : never;
  │ lhs (normalized) = Function#0;
  │ Member as {
  │     apply(this: Function, thisArg: any, argArray?: any): any;
  │     call(this: Function, thisArg: any, ...argArray: any[]): any;
  │     bind(this: Function, thisArg: any, ...argArray: any[]): any;
  │     toString(): string;
  │     prototype: any;
  │     readonly length: number;
  │     arguments: any;
  │     caller: Function;
  │ };
  │ rhs (normalized) = T#14#0 extends Function#0 ? T : never; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:587:73)
  │ context: tried to assign the true type of a conditional type to lhs (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:1089:22)
  │ context:
  │ lhs = Function#0;
  │ Member as {
  │     apply(this: Function, thisArg: any, argArray?: any): any;
  │     call(this: Function, thisArg: any, ...argArray: any[]): any;
  │     bind(this: Function, thisArg: any, ...argArray: any[]): any;
  │     toString(): string;
  │     prototype: any;
  │     readonly length: number;
  │     arguments: any;
  │     caller: Function;
  │ };
  │ rhs = T#14#0;
  │ lhs (normalized) = Function#0;
  │ Member as {
  │     apply(this: Function, thisArg: any, argArray?: any): any;
  │     call(this: Function, thisArg: any, ...argArray: any[]): any;
  │     bind(this: Function, thisArg: any, ...argArray: any[]): any;
  │     toString(): string;
  │     prototype: any;
  │     readonly length: number;
  │     arguments: any;
  │     caller: Function;
  │ };
  │ rhs (normalized) = T#14#0; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:587:73)
  │ context: LHS (final): Function#0;
  │ Member as {
  │     apply(this: Function, thisArg: any, argArray?: any): any;
  │     call(this: Function, thisArg: any, ...argArray: any[]): any;
  │     bind(this: Function, thisArg: any, ...argArray: any[]): any;
  │     toString(): string;
  │     prototype: any;
  │     readonly length: number;
  │     arguments: any;
  │     caller: Function;
  │ };
  │ RHS (final): T#14#0; (at crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs:1694:25)
  │ WrongArgType {
  │     span: Span {
  │         lo: BytePos(
  │             1115,
  │         ),
  │         hi: BytePos(
  │             1120,
  │         ),
  │         ctxt: #14,
  │     },
  │     inner: AssignFailed {
  │         span: Span {
  │             lo: BytePos(
  │                 1115,
  │             ),
  │             hi: BytePos(
  │                 1120,
  │             ),
  │             ctxt: #14,
  │         },
  │         cause: [],
  │     },
  │ }
    ╭─[./a.ts:47:1]
 47 │         toString1(value);
 48 │         toString2(value);
    ·                   ─────
 49 │     }
    ╰────
```
but I suppose that's because my solution isn't entirely correct? I can investigate more if we have to use `with_ctx`

kdy1: If so, I think it's one of

 - You didn't use `with_ctx` in some code. You have to wrap all operations on `self` with it.
 - It passes wrongly while it should not, because this change makes the Analyzer stateful.

Feel free to ignore it again.

kdy1: You should modify https://github.com/dudykr/stc/blob/e28c20fd9fc24940e47d7b8d319ce1847ba626c8/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs#L3240 and https://github.com/dudykr/stc/blob/e28c20fd9fc24940e47d7b8d319ce1847ba626c8/crates/stc_ts_file_analyzer/src/analyzer/expr/call_new.rs#L3250

If it still fails when you change both, it means the latter is the case.

awareness481: It worked by adding `with_ctx` on the second match arm. Im still not 100% sure my solution is good enough so please take that into account.

kdy1 (approved):
Thank you!

PR #728, opened 2023-02-18
fix(ts/analyzer): allow access interface in module

togami2864:
**Description:**
This PR allows to access interface type in the module, frequently used in jsx syntax.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #729, opened 2023-02-18
chore: update debugger config

togami2864:
**Description:**
update debug file to run tests with debugger under the same condition as conformance tests.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

PR #736, opened 2023-02-21
Fix: this binding at method assign function 

sunrabbit123:
**Description:**
```ts
impl.explicitVoid1 = function () {
  return 12;
};
impl.explicitVoid2 = () => 12;
impl.explicitStructural = function () {
  return this.a;
};
impl.explicitInterface = function () {
  return this.a;
};
impl.explicitStructural = () => 12;
impl.explicitInterface = () => 12;
impl.explicitThis = function () {
  return this.a;
};
```


crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -433,32 +433,40 @@ impl Analyzer<'_, '_> {
...
+                {
+                    id.sym == js_word!("this")
+                } else {
+                    false
+                }
+            } else {
+                false
+            };
+
+            let rhs_is_arrow = matches!(e.right, box RExpr::Arrow(RArrowExpr { .. }));
```
kdy1, line 456: Same here. I think we should handle `(() => {})` too

kdy1 (changes requested):
Thank you!
I think dealing directly with AST is hard because of this kinds of issues

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -433,32 +433,40 @@ impl Analyzer<'_, '_> {
             is_valid_lhs(&e.left).report(&mut analyzer.storage);
 
             let mut errors = Errors::default();
-            let will_skip_this = if let Some(ty) = type_ann {
+            // skip this flag for arrow function, static method etc.
+            let mut left_function_declare_not_this_type = false;
+            let right_function_declared_this = if let box RExpr::Fn(RFnExpr {
```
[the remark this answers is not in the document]
sunrabbit123: thx.

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
sunrabbit123: I fixed it [`2d2473f` (#736)](https://github.com/dudykr/stc/pull/736/commits/2d2473f63f3026baf7fe6e50ceaf834818ed801b)

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -484,13 +475,56 @@ impl Analyzer<'_, '_> {
                     let mut analyzer = analyzer.with_ctx(ctx);
 
                     let result = match type_ann {
-                        Some(ty) if will_skip_this => {
+                        Some(ty) if rhs_is_arrow || !right_function_declared_this => {
+                            let mut ty = ty.clone();
+                            if lhs_declared_this {
+                                if let Type::Function(stc_ts_types::Function { params, .. }) = ty.normalize_mut() {
```
kdy1, line 481: Please use `if let Some(stc_ts_types::Function { params, .. }) = ty.as_fn_type_mut()`.

kdy1: This un-freeze `ty` only if `ty` is `Type::Function`

kdy1 (changes requested):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
sunrabbit123, line 481:
```suggestion
                                if let Some(stc_ts_types::Function { params, .. }) = ty.as_fn_type_mut() {
```

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -484,13 +475,56 @@ impl Analyzer<'_, '_> {
...
+                                }
+                            }
+
+                            e.right
+                                .validate_with_args(&mut *analyzer, (mode, None, Some(&ty)))
+                                .context("tried to validate rhs an assign expr")
+                        }
+                        Some(ty) if right_function_declared_this => {
                             let mut ty = ty.clone();
                             if let Type::Function(stc_ts_types::Function { params, .. }) = ty.normalize_mut() {
```
sunrabbit123, line 511:
```suggestion
                            if let Some(stc_ts_types::Function { params, .. }) = ty.as_fn_type_mut() {
```

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
sunrabbit123: i fixed it

kdy1 (approved):
Thank you so much!

PR #747, opened 2023-02-23
Fix: assign conditional type with function and intersection

sunrabbit123:
**Description:**

```ts
type Foo = { foo: string };
type Bar = { bar: string };

declare function fooBar(x: { foo: string; bar: string }): void;
declare function fooBat(x: { foo: string; bat: string }): void;
type Extract2<T, U, V> = T extends U ? (T extends V ? T : never) : never;

function f20<T>(
  x: Extract<Extract<T, Foo>, Bar>,
  y: Extract<T, Foo & Bar>,
  z: Extract2<T, Foo, Bar>
) {
   fooBar(x);
   fooBar(y);
   fooBar(z);
}
```


crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -1085,7 +1085,13 @@ impl Analyzer<'_, '_> {
 
         match (to, rhs) {
             (_, Type::Conditional(rc)) => {
-                self.assign_with_opts(data, to, &rc.true_type, opts)
+                let ty = if rc.true_type.type_eq(&rc.check_type) {
```
kdy1, line 1088: Is this logic correct? Not sure why is `Foo extends Bar ? Foo : Baz` special

sunrabbit123: this pr is incomplete

reason is simple
```ts
type Bar = string;
function Test1<T extends Bar>(arg1 : T){	
	const some : Bar = arg1; // ok
  arg1 = "123" as Bar; // error
}
```

sunrabbit123: additional..

if check_type same true_type, true_type has extends_type
because check_type extends extends_type

1. check_type == true_type
2. check_type = extends_type
3. true_type = extends_type

sunrabbit123: Of course, I'd like to visit the lower type whistle,
but I forgot the name of the function name.

sunrabbit123:
I think that this pr is too early.
I think the intersection work should be made clear first.

crates/stc_ts_type_checker/tests/conformance/controlFlow/controlFlowGenericTypes.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
     required_error: 6,
     matched_error: 2,
-    extra_error: 12,
+    extra_error: 13,
```
kdy1, line 4: Seems like a regression?

sunrabbit123: he need exclude_types_using_fact, but he not has name

sunrabbit123: The result is a false type inference.

sunrabbit123: I think that typefact about truty add logic
`NonNullable`

kdy1 (changes requested):
Thanks!

crates/stc_ts_type_checker/tests/conformance/jsx/tsxAttributeResolution5.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
     required_error: 3,
     matched_error: 0,
-    extra_error: 0,
+    extra_error: 1,
```
sunrabbit123, line 4: The error about err_shift_n

crates/stc_ts_file_analyzer/src/analyzer/types/conditional.rs
```diff
@@ -0,0 +1,283 @@
...
+    VResult,
+};
+
+impl Analyzer<'_, '_> {
+    pub(crate) fn normalize_conditional<'a>(
+        &mut self,
+        actual_span: Span,
+        c: &Conditional,
+        opts: NormalizeTypeOpts,
+    ) -> VResult<Cow<'a, Type>> {
```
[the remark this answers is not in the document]
sunrabbit123: this function is refer and recursive `normalize`
It is needed

crates/stc_ts_types/src/lib.rs
```diff
@@ -1611,15 +1617,44 @@ impl Type {
         }
         tys.dedup_type();
```
kdy1, line 1619: I think we should remove this logic instead of expanding

sunrabbit123: all? or a bit?

I think we need things that look like the next code.
```rs
        if tys.iter().any(|ty| ty.is_never()) {
            return Type::never(span, Default::default());
        }
```

sunrabbit123: Instead, I don't think I need the rest.

kdy1: I think we should rely on `normalize_intersection_types` instead of converting it to `never` on creation.

sunrabbit123: but, I think that it need simple check for never.
That's how the original chord works.

kdy1: If so, can you extract this as `fn is_trivial_never(types: &[Type]) -> bool` of `Intersection`?

```rust
impl Intersection {
    pub fn fn is_trivial_never(types: &[Type]) -> bool {
    }
}
```

sunrabbit123: Good. I think that's a good idea.
Is this the code?

kdy1: It should filled with logic in `new_intersection`

sunrabbit123: What kind of logic?

I put in what I accidentally left out.
I broke the test because of them :D

cspell.json
```diff
@@ -90,6 +90,7 @@
     "tempdir",
     "testfn",
     "tikv",
+    "trival",
```
kdy1, line 93: This looks like `trivial`

kdy1: Is it my typo? Sorry 🤣

crates/stc_ts_types/src/lib.rs
```diff
@@ -1482,6 +1488,57 @@ impl Intersection {
 
         self.visit_with(&mut AssertValid);
     }
+
+    pub fn is_trival_never(iter: &[Type]) -> bool {
```
kdy1, line 1492: Please rename this to `is_trivial_never`

sunrabbit123, line 1492:
```suggestion
    pub fn is_trivial_never(iter: &[Type]) -> bool {
```

crates/stc_ts_file_analyzer/src/analyzer/generalize.rs
```diff
@@ -271,12 +271,8 @@ impl Fold<Type> for Simplifier<'_> {
             }
 
             Type::Intersection(ref i) => {
-                let is_str = i.types.iter().any(|ty| ty.is_str());
-                let is_num = i.types.iter().any(|ty| ty.is_num());
-                let is_bool = i.types.iter().any(|ty| ty.is_bool());
-
                 // LHS is never.
-                if u32::from(is_str) + u32::from(is_num) + u32::from(is_bool) >= 2 {
+                if Intersection::is_trival_never(&i.types) {
```
sunrabbit123, line 275:
```suggestion
                if Intersection::is_trivial_never(&i.types) {
```

cspell.json
sunrabbit123, line 93:
```suggestion
```

crates/stc_ts_types/src/lib.rs
```diff
@@ -1656,15 +1713,7 @@ impl Type {
...
-        let has_str = tys.iter().any(|ty| ty.is_str());
-        let has_bool = tys.iter().any(|ty| ty.is_bool());
-        let has_num = tys.iter().any(|ty| ty.is_num());
-
-        if u32::from(has_str) + u32::from(has_bool) + u32::from(has_num) > 1 {
-            return Type::never(span, Default::default());
-        }
-
-        if tys.iter().any(|ty| ty.is_never()) {
+        if Intersection::is_trival_never(&tys) {
```
sunrabbit123, line 1716:
```suggestion
        if Intersection::is_trivial_never(&tys) {
```

cspell.json
sunrabbit123: 😢 It my typo and misunderstand
I fixed it

crates/stc_ts_types/src/lib.rs
```diff
@@ -1673,6 +1722,14 @@ impl Type {
             tys.retain(|ty| !ty.is_unknown());
         }
 
+        if tys.len() > 1
+            && !(tys.len() == 2 && tys.iter().any(|ty| ty.is_type_param() || ty.is_conditional()))
+            && tys.iter().any(|ty| ty.is_type_lit())
+        {
+            // reduce empty type lit
+            tys.retain(|ty| if let Type::TypeLit(ty) = ty { !ty.is_empty() } else { true });
```
kdy1, line 1730:
```suggestion
            tys.retain(|ty| if let Type::TypeLit(ty) = ty.normalize() { !ty.is_empty() } else { true });
```

crates/stc_ts_type_checker/tests/tsc-stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 3866,
-    matched_error: 6135,
-    extra_error: 867,
-    panic: 105,
+    required_error: 3784,
+    matched_error: 6027,
```
kdy1, line 3: Please rebase correctly. Seems like CI is not catching this properly

kdy1 (changes requested):

crates/stc_ts_type_checker/tests/conformance/jsx/tsxAttributeResolution5.stats.rust-debug
kdy1: I don't think this is a problem of `err_shift_n`

kdy1: This is actually a regression.
I checked it

kdy1: But well, let's ignore this for now

kdy1 (approved):
Thanks!

PR #751, opened 2023-02-23
feat: Respect type constraints while inferring types

kdy1:
**Description:**

 - Abort on constraint mismatch.
 - Don't report argument type mismatch if inference fails.
 - Implement more type inference rules for template literals.


kdy1 (commented):
I'll merge this with regressions as all the fixes for regressions should be separate PRs.

PR #752, opened 2023-02-23
fix(ts/analyzer): return Promise<any> when `import` got string union as an arg

togami2864:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #759, opened 2023-02-25
implement more of is_type_related_to

awareness481:
**Description:**

`type_related_to`, will write more when ready

Todo:

- [ ] Figure out regressions


awareness481:
Will decide soon if I can continue or not with this PR, facing some issues

awareness481:
@kdy1 Spent a lot of time trying out things but I ended up on the conclusion that this is not worth implementing until `ty.is_substitution()` is implemented (and that one seems pretty hard to implement), as otherwise it would need hacky logic.

With this PR I was trying to implement the following part of the `isTypeRelatedTo` function

```ts
        if (relation !== identityRelation) {
            if (relation === comparableRelation && !(target.flags & TypeFlags.Never) && isSimpleTypeRelatedTo(target, source, relation) || isSimpleTypeRelatedTo(source, target, relation)) {
                return true;
            }
        }
        else if (!((source.flags | target.flags) & (TypeFlags.UnionOrIntersection | TypeFlags.IndexedAccess | TypeFlags.Conditional | TypeFlags.Substitution))) {
            // We have excluded types that may simplify to other forms, so types must have identical flags
            if (source.flags !== target.flags) return false;
            if (source.flags & TypeFlags.Singleton) return true;
        }
```

kdy1:
Ah, I understand. Thank you anyway for trying/investigating

PR #760, opened 2023-02-26
fix: Improve selection of `TS2322` vs `TS2741`

kdy1:
**Description:**

This PR introduces `report_assign_failure_for_missing_properties` in `AssignOpts` which can be used to select between `TS2322` and `TS2471` for the caller side.


kdy1:
I stopped working on this PR because we need other tasks to continue working on assignments. (https://github.com/dudykr/stc/issues/761, https://github.com/dudykr/stc/issues/762)

PR #764, opened 2023-02-27
fix: passing type_param to Function without type_param

awareness481:
**Description:**

Fixes the required errors on these 3 lines

https://github.com/awareness481/stc/blob/528742ae6a6d684eb3f9e0502899d2c623c839f5/crates/stc_ts_type_checker/tests/conformance/expressions/functionCalls/functionCalls.ts#L33-L35


kdy1 (approved):
Thank you!

PR #765, opened 2023-02-27
handle intrinsic uppercase/lowercase/capitalize/uncapitalize string inputs

awareness481:
**Description:**

Handles most test cases related to the intrinsic string types (Uppercase/Lowercase/Uncapitalize/Capitalize).  AFAICT most of the not solved ones are caused by https://github.com/dudykr/stc/issues/771 and some are a bit more advanced,


crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2808,13 +2808,94 @@ impl Analyzer<'_, '_> {
     /// Should be called only if `to` is not expandable.
```
awareness481, line 2808: Not sure about this, I was expecting to get errors (since im not checking if to is not expandandable) but I didn't get any. Any pointers would be helpful

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2808,13 +2808,94 @@ impl Analyzer<'_, '_> {
...
     fn assign_to_intrinsic(&mut self, data: &mut AssignData, to: &StringMapping, r: &Type, opts: AssignOpts) -> VResult<()> {
         match to.kind {
-            IntrinsicKind::Uppercase => {}
-            IntrinsicKind::Lowercase => {}
-            IntrinsicKind::Capitalize => {}
-            IntrinsicKind::Uncapitalize => {}
+            IntrinsicKind::Uppercase => {
```
awareness481, line 2811: Generally I think the code is overly verbose for what's its doing, any suggestions?

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2808,13 +2808,94 @@ impl Analyzer<'_, '_> {
...
+                                    right_ident: None,
+                                    right: box r.clone(),
+                                    cause: vec![],
+                                }
+                                .into());
+                            }
+                        }
+                    }
+                }
+            }
```
awareness481, line 2894: Generally I think the code is overly verbose for what's its doing, any suggestions?

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: I also tried
```rs
let a = &self.expand(r.span(), Type::StringMapping(to.clone()), Default::default());

        if let Ok(Type::StringMapping(var)) = a {
            match to.kind {
```
and that behaved the same way so I ended up not using it because it seems wasteful. I tried to find an example where not checking whether `to` is expandable would be an issue but I couldnt (for now the function `.assign_to_intrinsic` is only called in one place, perhaps if it is called in more it can become an issue?).

awareness481:
Switching back to draft because I haven't checked if this will work well with the other required errors for intrisinc string types

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: code will likely change so nvm this

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: Based on the comment here https://github.com/dudykr/stc/blob/af5fda3c18a132ee7959ffb236bfc1879f41951d/crates/stc_ts_type_checker/tests/conformance/types/literal/stringMappingOverPatternLiterals.ts#L74-L79 I think we actually don't want to expand the intrinsic type

kdy1: The comment is not about an error. It means you are not expected to call it if `to` is expandable.

kdy1: And sorry for late response, I missed it.

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2812,14 +2812,168 @@ impl Analyzer<'_, '_> {
...
+                                right: box r.clone(),
+                                cause: vec![],
+                            }
+                            .into());
+                        }
+                    }
+                }
+                IntrinsicKind::Capitalize => {
+                    if let Some(value) = &value.raw {
+                        let ch = value.to_string().chars().next();
```
kdy1, line 2849: `.to_string()` seems redundant

kdy1:
```suggestion
                        let ch = value.chars().next();
```

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2812,14 +2812,168 @@ impl Analyzer<'_, '_> {
...
+                                    cause: vec![],
+                                }
+                                .into());
+                            }
+                        }
+                    }
+                }
+                IntrinsicKind::Uncapitalize => {
+                    if let Some(value) = &value.raw {
+                        let ch = value.to_string().chars().next();
```
kdy1, line 2867: Same here

kdy1:
```suggestion
                        let ch = value.chars().next();
```

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2812,14 +2812,168 @@ impl Analyzer<'_, '_> {
...
+                        cause: vec![],
+                    }
+                    .into());
+                }
+
+                return Ok(());
+            }
+            Type::Tpl(tpl) => match to.kind {
+                IntrinsicKind::Uppercase => {
+                    let is_not_uppercase = tpl.quasis.iter().all(|s| s.value == s.value.to_uppercase());
```
kdy1, line 2900: You can do `s.value.is_uppercase()`

kdy1: And I think the variable name is confusing

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2812,14 +2812,168 @@ impl Analyzer<'_, '_> {
...
+                            right: box r.clone(),
+                            cause: vec![],
+                        }
+                        .into());
+                    }
+
+                    return Ok(());
+                }
+                IntrinsicKind::Lowercase => {
+                    let is_not_uppercase = tpl.quasis.iter().all(|s| s.value == s.value.to_lowercase());
```
kdy1, line 2916: Same here, you can use `is_lowercase()`

kdy1 (changes requested):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: I dont think that I can do `s.value.is_uppercase` because `s.value` is an atom but let me know if Im missing something. In any case I want to re-verify the logic for template literals so please let me do that and I'll ask for a review again :pray:

awareness481: NVM, Im positive that `Uppercase<tpltype>` and `Lowercase<tpltype>` are affected by https://github.com/dudykr/stc/issues/771 so I ll just correct the if conditions for now

kdy1: See: https://doc.rust-lang.org/book/ch15-02-deref.html

The methods of `&str` can be used for atom types, too.

awareness481: Oh I asked for review before I saw this comment, I ll do that when I get back, sorry for the noise!

awareness481: @kdy1 I apologize if I'm wrong but I don't think `&str` has `.is_upperacase()` only  `to_uppercase()`.... right? :sweat_smile:

kdy1: If so you can use .chars().all(|c| ...)

awareness481: done :P

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2812,14 +2812,168 @@ impl Analyzer<'_, '_> {
...
-            IntrinsicKind::Lowercase => {}
-            IntrinsicKind::Capitalize => {}
-            IntrinsicKind::Uncapitalize => {}
+        match r {
+            Type::Lit(LitType {
+                lit: RTsLit::Str(value), ..
+            }) => match to.kind {
+                IntrinsicKind::Uppercase => {
+                    if let Some(value) = &value.raw {
+                        if !value.to_uppercase().eq(&value.to_string()) {
```
kdy1, line 2821: `to_string()` seems redundant, and `==` is better than `.eq()`

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2812,14 +2812,168 @@ impl Analyzer<'_, '_> {
...
+                                right: box r.clone(),
+                                cause: vec![],
+                            }
+                            .into());
+                        }
+                    }
+                }
+                IntrinsicKind::Lowercase => {
+                    if let Some(value) = &value.raw {
+                        if !value.to_lowercase().eq(&value.to_string()) {
```
kdy1, line 2835: Same here. `JdWord` implements `Deref<Target = str>` so you can do `&**value` to get `&str`

kdy1 (changes requested):
Thanks!

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: done,  clippy made me do

```diff
- &**value
+ **value
```
though, I hope that's ok

kdy1 (approved):

kdy1 (dismissed):

kdy1 (approved):

Issue #720, opened 2023-02-16
Default type for an assignment patterns should respect all values


kdy1:
https://github.com/dudykr/stc/blob/1cb6c1844cfd2e6b069bddc3588dda51d53a3b6f/crates/stc_ts_type_checker/tests/conformance/es6/destructuring/destructuringWithLiteralInitializers2.ts#L15

Currently `stc` infers the type as `number[]`, but it should be `[number, (string | undefined)]`

blacktoast:
Could I try to fix it?

kdy1:
Of course, thank you!

PR #773, opened 2023-03-01
feat: try fix

blacktoast:
**Description:**
```

  x Type
    ,-[$DIR/tests/pass/es6/destructuring/destructuringWithLiteralInitializers2/1.ts:16:1]
 16 | function f03([x, y, z] = [1]) {}
 17 | f03();
    : ^^^
    `----

Error: 
  > ([x, y, z]?: ((number | any))[]) => void
```
Fixed an issue where the type was inferred to be an array so that it could be inferred as follows.
```
Error: 
  > [1]

  x Type
    ,-[$DIR/tests/pass/es6/destructuring/destructuringWithLiteralInitializers2/1.ts:13:1]
 13 | function f02([x, y] = [1]) {}
 14 | f02();
    : ^^^
    `----
Error: 
  > ([x, y]?: [number, any]) => void
```

I modified the `add_vars` function to do this.
Added the any type for unassigned values when declaring functions.

This fix makes a panic bug, but I'll fix it soon.

I'm wondering how to change the `any` type to `any?`

**BREAKING CHANGE:**

#720


blacktoast:
This is not complete, but can you please help check if I'm in the right direction?

kdy1:
For any => any?, you can create a `Type::Optional`

crates/stc_ts_file_analyzer/src/analyzer/scope/vars.rs
```diff
@@ -160,20 +172,58 @@ impl Analyzer<'_, '_> {
...
                 if self.ctx.is_fn_param && type_ann.is_none() {
                     // If the declaration includes an initializer expression (which is permitted
                     // only when the parameter list occurs in conjunction with a
                     // function body), the parameter type is the widened form (section
                     // 3.11) of the type of the initializer expression.
+                    match &*p.left {
+                        RPat::Array(p_left) => {
+                            println!("tuple@@@!!! {:?}", right);
+
+                            right = right.fold_with(&mut Widen { tuple_to_array: false });
```
kdy1, line 196: The direction looks fine, but isn't this logic common for all cases?
I think this can be moved to up of the match expression

blacktoast: In my opinion, `right = right.fold_with(&mut Widen { tuple_to_array: false });` is not common logic.

` right = right.fold_with(&mut Widen { tuple_to_array: ture });` is the existing logic, 
If I do this In situations like `function f03([x, y, z] = [1])`, the type is inferred as `number[]` instead of `[number]`.

So if left is an array, and `is_fn_param=true`, I want to do the logic `right = right.fold_with(&mut Widen { tuple_to_array: false });`,

kdy1: Ah, my bad. I missed the `tuple_to_array`

blacktoast:
I've fixed the issue #720

However, another issue has arisen.

I modified the `add_vars` function to change the type for each function parameter to fix the type inference

but The problem is :
```ts
declare const nx: undefined;
declare const sx: string;

function f33([x = 0, y = sx] = [1]) {
  y;
  x;
}
f33(); 
```

In the following code: f33() is correctly inferred to be`([x = 0, y = sx]?: [number, (string | undefined)?]) => void', but inside the function, y is also inferred as `((string | undefined)? | string)`.

To be honest, I don't know what to fix.
How can I fix it?

kdy1:
You should modify the logic in `add_var`. I think adding some `dbg!` to `declare_var` will help

crates/stc_ts_type_checker/tests/tsc-stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
```
kdy1, line 1: CI failed

kdy1 (changes requested):

PR #783, opened 2023-03-04
fix(ts/analyzer): disallow the use of specific keyword in indexed access

togami2864:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #797, opened 2023-03-06
Feature: Fix 2322 about typeof this

sunrabbit123:
**Description:**
```ts
const Test8 = () => {
  let x: typeof this.no = 1;
};
```


kdy1 (approved):
Thanks!

PR #798, opened 2023-03-06
feat: Create cond facts for RHS on `&&=`

### LINKED ISSUE #595: `&&=` should create type fact for the RHS
https://github.com/dudykr/stc/blob/273356f921c6152b5a45c41527923a86361ebe13/crates/stc_ts_type_checker/tests/conformance/es2021/logicalAssignment/logicalAssignment4.ts#L28-L33

In the code above, `thing &&= thing.original` should pass, but stc fails to validate it

hwoongkang:
**Description:**

1. About the FIx.
If we are to validate rhs in a `&&=` statement, it implies that the variable in lhs is truthy. But the problem was that the typefact is propagated outside the`if` test (control flow), but was not being applied in the validation process of rhs.  
I first tried to apply the fact directly, but it resulted in breaking the other logicalAssignment tests (5-8).  
I find my own construction of `CondFacts` not so pretty, it would be great if there is another way to build a child scope.

2. About the Remaining Errors.

```Typescript
} else if (v === 3) {
    if ((thing ||= defaultValue)) {
      thing.name;
      defaultValue.name;
    }
```

The code above should emit error, but currently, stc fails to do that.  
My fix only removes the extra error in the lines below:

```Typescript
  if (v === 1) {
    if ((thing &&= thing.original)) {
      thing.name;
    }
```

As I was examining the remaining issue, I think that extra errors are related to the code block below:

https://github.com/dudykr/stc/blob/a65025337e34695ef44f7a6153e23c6979cbfe8a/crates/stc_ts_file_analyzer/src/analyzer/control_flow.rs#L349-L370

As I examined the `facts_from_test` variable, it has the type facts that `defaultValue` is `TypeFacts::NEUndefined | TypeFacts::NENull | TypeFacts::NEUndefinedOrNull | TypeFacts::Truthy`, which is not true.

I think validation of logical OR expression is also broken.

**Related issue:**

 - Closes https://github.com/dudykr/stc/issues/595.


crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -516,9 +527,17 @@ impl Analyzer<'_, '_> {
                                 }
                             }
 
-                            e.right
-                                .validate_with_args(&mut *analyzer, (mode, None, Some(&ty)))
-                                .context("tried to validate rhs an assign expr")
+                            if let Some(facts) = with_child_scope {
+                                analyzer.with_child(ScopeKind::Block, facts, |analyzer| {
```
kdy1, line 531: I think this should be something different than `ScopeKind::Block`.

kdy1: Maybe `ScopeKind::ControlFlow`?

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -366,6 +367,7 @@ impl Analyzer<'_, '_> {
 
             let mut left_i = None;
             let ty_of_left;
+            let mut with_child_scope: Option<CondFacts> = None;
```
kdy1, line 370: Can you rename this to `facts_for_rhs`?

kdy1 (changes requested):
Thank you!

Can you sign the CLA?

hwoongkang:
Signed the CLA!

OK I will apply the changes requested in a moment. Thank you

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
hwoongkang: couldn't find `ScopeKind::ControlFlow`, maybe you meant `ScopeKind::Flow`?

kdy1 (changes requested):
I found that this approach is wrong because we should preserve type facts for consequent expressions

hwoongkang:
Oh OK.

I am not quite sure which are the type facts that you said should be preserved, but here is my observation:

In the validation process of `RIfStmt` in the `control_flow.rs` file I linked in the first comment, both of `thing` and `thing.original` do have `Truthy` type facts.

Thus my thoughts were that, the assignment express itself is correctly working (and "preserving" the facts from the assignment expression), and the problem lies in the validation step of rhs.

That's why I tried to make a subtree, or a child scope, to make rhs validation work, while minimizing the side effects outside the scope.

--

If I am understanding the problem completely wrong, please let me know.

And thanks for the review :)

kdy1:
The problem is that your logic is too specific to input.

```ts
declare var a: string | null
if (!a) {
    a ??= "foo"
}

console.log(a)
```

```ts
declare var b: { p: string | null }
if (!b.p) {
    b.p ??= "foo"
}

console.log(b.p)
```

hwoongkang:
IMHO, I still don't get it, since yout example is about `??=` and the issue and my fix was about `&&=`.

I guess my understandings about the codebase is still lacking. might try later with another approach. Thanks!

kdy1:
Ah, you can change it to `&&=`

kdy1:
Oh, my bad.
I found that there's a logic for updating `actual_ty`.
I'll rebase/merge this PR.

kdy1 (approved):

Issue #771, opened 2023-02-28
stc fails to find Intrinsic string type in some cases


awareness481:
Given the following code

```ts
type A = Uppercase<42>;
```

if you run `stc` on it you'll notice that stc does not report an error. Upon further investigation it looks like `stc` fails to find the built-in type `Uppercase` in this case. I think it is because `self.config.builtin` is set to `false`.

Note that it only fails to find the type in some cases, in most cases (like then `Uppercase` is a type of a function param, stc finds the type)

awareness481:
As far as I know this should always be an error

```ts
type Uppercase = string;
//   ?^ Duplicate identifier 'Uppercase'.
```

**unless** https://www.typescriptlang.org/tsconfig#noLib is enabled. So perhaps something like that should work for all TS builtins? 

```
type Uppercase = string;
//   ?^ no error, noLib: true
```

If `noLib` is out of scope for `0.0.1` it could always be assumed that defining a type with has the same identifier as a TS builtin is an error

kdy1:
I'm not sure about `noLib`, but I think we can exclude it for now.
I don't expect someone to use it for a real-world project

awareness481:
Completely agree with `noLib` being excluded, but as far as the issue goes, I think I got confused because I was using `type Uppercase = string` to be debug, but the `type` keyword is not implemented yet right? Which is why it said `Uppercase missing type`.

I think I should close the issue.

kdy1:
Type aliases are implemented

kdy1:
But I'm confused too. This is about `Uppercase<42>`, right?
`type Uppercase = string` seems not relevant

awareness481:
> But I'm confused too. This is about `Uppercase<42>`, right? `type Uppercase = string` seems not relevant

Yes, but in that case it fails to find `type A` and so it doesn't check `Uppercase<42>` AFAICT

awareness481:
> ```ts
> type A = Uppercase<42>;
> ```

Maybe it's a bug with type aliases then, here's the log with the only `type A = Uppercase<42>;`

```rs
  DEBUG [(0)/types] Registering: A#2
    at crates/stc_ts_file_analyzer/src/analyzer/scope/mod.rs:791
    in register_type
    in Stmt with line_col: "(33:1-33:24)"

  DEBUG (0) Scope.find_type(`A#2`)
    at crates/stc_ts_file_analyzer/src/analyzer/scope/mod.rs:1192
    in find_local_type with name: A#2
[...]

PR #800, opened 2023-03-06
type-checking string intrinsic types in type aliases

awareness481:
**Description:**

The rhs of type aliases, is not type-checked (not sure yet if it's for string intrinsic types only or all types). For example, the following would pass

```ts
type A = Uppercase<42>`
```
even tho `Uppercase<42>` is an invalid type. This PR tries to fix that and all the regressions that come from type-checking the rhs (that was previously not type-checked)

closes https://github.com/dudykr/stc/issues/771


crates/stc_ts_file_analyzer/src/analyzer/convert/mod.rs
```diff
@@ -225,7 +224,34 @@ impl Analyzer<'_, '_> {
                         metadata: Default::default(),
                     }),
 
-                    _ => d.type_ann.validate_with(child)?,
+                    t => {
+                        if let RTsType::TsTypeRef(type_ref) = t {
+                            if let RTsEntityName::Ident(t) = &type_ref.type_name {
+                                match &*t.sym {
+                                    "Uppercase" | "Lowercase" | "Capitalize" | "Uncapitalize" => {
```
awareness481, line 231: String intrinsic types in type alias assignments aren't caught by the https://github.com/dudykr/stc/blob/7bfe8ea0a4b4407f6517c8978ecd3d04de7afdb8/crates/stc_ts_file_analyzer/src/analyzer/convert/mod.rs#L199-L201
 match arm so thats why we're doing this

crates/stc_ts_type_checker/tests/conformance/types/literal/stringMappingOverPatternLiterals.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 2,
-    matched_error: 27,
+    required_error: 6,
+    matched_error: 23,
```
awareness481, line 3: Regression here is because I haven't implemented a way yet to figure out whether we're in a variable assignment

```ts
let x: Uppercase<string>;
let y: Lowercase<string> = 'lower'
x = y // should be error
```
or a type alias assignment
```ts
type A = Uppercase<string> // should be no error
```
(and I think I might need to differentiate for type params, but not 100% sure yet)
```ts
type X<T extends string> = Uppercase<T>;
type S = X<Lowercase<string>> // no error
```

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2806,8 +2832,58 @@ impl Analyzer<'_, '_> {
...
+                                    .into());
+                                }
+                            }
+                        }
+                    }
                 }
-            },
+            }
+            Type::Union(union) => {
+                if union.types.iter().all(|v| v.is_str_like()) {
```
awareness481, line 2874: Need to handle union better

kdy1: Nested unions are not allowed, so it may be enough

crates/stc_ts_type_checker/tests/conformance/types/literal/stringMappingOverPatternLiterals.stats.rust-debug
awareness481: Ended up doing https://github.com/dudykr/stc/blob/6fd7e1db7630dff38f0f8d4e5d5badee4ec15250/crates/stc_ts_file_analyzer/src/analyzer/convert/mod.rs#L284-L292

to differentiate. Not sure if `in_actual_type` is the correct choice but I want to avoid creating another property without being sure this is the correct approach.

kdy1: I think validating an alias only if it's a specific type is a bad idea.

kdy1: I think it should instead be 

```rust
self.normalize(span, alias, Default::default).context("tried to normalize type alias").report(&mut self.stroage);
```

awareness481: FWIW I tried this, but it doesn't report any errors for type aliases :cold_sweat:

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -3028,7 +3028,23 @@ impl Analyzer<'_, '_> {
 
                 return Ok(());
             }
-
+            Type::Param(param) => {
+                if let Some(constraint) = &param.constraint {
+                    if !constraint.is_str_like() && !constraint.is_never() {
+                        return Err(ErrorKind::NotSatisfyConstraint {
+                            span: param.span,
```
awareness481, line 3035: span is wrong here, currently it's
```rs
 175 │ type X<T extends number> = Uppercase<T>;
     ·        ────────────────
```
but it needs to be
```rs
```rs
 175 │ type X<T extends number> = Uppercase<T>;
     ·                                      ─
```
looking into it
```

File: crates/stc_ts_type_checker/tests/conformance/types/literal/stringMappingOverPatternLiterals.stats.rust-debug (hunk 2 above)
Comment (kdy1) [reply crates/stc_ts_type_checker/tests/conformance/types/literal/stringMappingOverPatternLiterals.stats.rust-debug]: If so, I think you should add logic to https://github.com/dudykr/stc/blob/bfeb6fbb6211618e1165eca7710186d6427b2c8e/crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs#L1964-L2112

Comment (awareness481) [reply crates/stc_ts_type_checker/tests/conformance/types/literal/stringMappingOverPatternLiterals.stats.rust-debug]: I might be missing something, but I dont think this approach will work because I'm currently relying on `ctx` which doesn't seem to be passed on `self.normalize`

ex:
(doing this because I need to know if its a type alias assignment or not)
```
        let ctx = Ctx {
            in_actual_type: true,
            ..self.ctx
        };

        self.with_ctx(ctx)
            .normalize(Some(d.span), Cow::Borrowed(&alias), Default::default())?;
            
        // self.ctx.in_actual_type -> false inside the expand_intrinsic_types function
```

What I had thought of initially was a generic function to validate the rhs of the type alias assignment, because this problem doesn't seem unique to string intrinsic types. Im not sure what that would like exactly.

**Review by awareness481 — COMMENTED** [review]
https://github.com/dudykr/stc/blob/cdb849cf637830d340cff12f2f76e3317c0b5b58/crates/stc_ts_file_analyzer/tests/pass/types/tpl/stringMappingOverPatternLiterals/.1.ts

(disabling pass test)

Blocking change, has to be fixed before PR is ready

Edit: It is a false positive because previously it wasn't type checked, I still think it is a good idea to come to a conclusion about this before the PR is evaluated

**Comment by kdy1** [comment]
I think disabling tests in this case is fine because those were not actually _passing_. They were simply not validated.

File: crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs (hunk 4 above)
Comment (awareness481) [reply crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs]: Fixed but still slightly wrong, it's
```rs
 │ type X<T extends number> = Uppercase<T>;
     ·                                 ───
```
but it needs to be
```rs
 │ type X<T extends number> = Uppercase<T>;
     ·                                  ─
```
IMO it's kinda ok because string intrinsics only take one param.

kdy1: I agree with you. I don't think it matters for the initial version

crates/stc_ts_type_checker/tests/conformance/types/literal/stringMappingOverPatternLiterals.stats.rust-debug
awareness481: Ended up with 

```rs
if let Err(e) = self.normalize(Some(d.span), Cow::Borrowed(&alias), Default::default()) {
            return Err(e);
        }

        let ctx = Ctx {
            in_actual_type: true,
            ..self.ctx
        };

        if let Type::StringMapping(ty) = alias.normalize() {
            if let Err(e) = self
                .with_ctx(ctx)
                .assign_to_intrinsic(&mut Default::default(), ty, &ty.type_args.params[0], Default::default())
            {
                return Err(e);
            }
        }
```

This is done because I cant pass ctx in `.normalize`. I considered creating `self.validate_rhs_of_alias` but I think at this point Idk if we need something like that.

Extra errors and panics due to calling `self.normalize`

kdy1: You can add a field to `NormalizeOpts`, or do `self.with_ctx(ctx).normalize(...)`

awareness481: I'm really not sure what to make of this PR, calling `assign_to_intrinsic` causes so many side effects I eventually gave up on making it work. I'm unsure whether that's caused by my code (partially it was I had to do some fixes), or whether having `assign_to_intrinsic` inside `normalize` is problematic.

Perhaps this PR should be closed/blocked until type-checking for type aliases is implemented in some way?

kdy1: Hmm.. I'll try implementing validation for type alias soon.

kdy1: I did it 😄

awareness481: Oh yeah thanks a lot for unblocking me :pray: , I've been working on top of your changes for last couple of hours. It seems likely that this will work out :crossed_fingers:

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -80,6 +80,7 @@ pub(crate) struct NormalizeTypeOpts {
     pub expand_enum_variant: bool,
 
     pub preserve_keyof: bool,
+    pub in_type_or_type_param: bool,
```
awareness481, line 83: Added this because I want to know if I'm inside a type or a type_param. For example, in an assignment `Lowercase<string>` isn't assignable to `Uppercase<string>`

```ts
declare var x: Uppercase<string>
declare var y: Lowercase<string>
x = y ❌
```

but it an type param it is valid
```ts
declare function a(x: Uppercase<Lowercase<string>>) {} ✅
```

kdy1: Makes sense. Thank you!

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -224,6 +225,23 @@ impl Analyzer<'_, '_> {
                             .expand_intrinsic_types(actual_span, i)
                             .context("tried to expand intrinsic type as a part of normalization")?;
 
+                        let ctx = Ctx {
+                            in_actual_type: opts.in_type_or_type_param,
+                            ..self.ctx
+                        };
+
+                        if let Type::StringMapping(ty) = ty.normalize() {
```
kdy1, line 233: Can you move this logic to `expand_intrinsic_types`?

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2785,47 +2787,67 @@ impl Analyzer<'_, '_> {
...
                     }
                 }
-                IntrinsicKind::Capitalize => {
-                    if let Some(value) = &value.raw {
-                        let ch = value.chars().next();
 
-                        if let Some(ch) = ch {
-                            if !ch.is_uppercase() {
+                // TODO: Handle this better
+                let mut span = opts.span;
```
kdy1, line 2825: `let span = opts.span.or_else(|| to.span())` will work.

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2785,47 +2788,74 @@ impl Analyzer<'_, '_> {
...
+            Type::Lit(LitType {
+                lit: RTsLit::Str(str_lit), ..
+            }) => {
+                let type_param = &to.type_args.params[0];
+
+                if self.ctx.in_actual_type {
+                    return Ok(());
+                }
+
+                if type_param.is_alias() || type_param.is_ref_type() {
```
kdy1, line 2849: This seems wrong. Shouldn't we normalize it instead?

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2883,61 +2931,208 @@ impl Analyzer<'_, '_> {
...
+                    right: box Type::Keyword(KeywordType {
+                        kind: ty,
+                        span: r.span(),
+                        metadata: Default::default(),
+                        tracker: Default::default(),
+                    }),
+                }
+                .into());
+            }
+            Type::Union(union) => {
```
kdy1, line 2953: Can you use another name instead of `union`? GitHub highlighter treats this as a keyword.

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2883,61 +2931,208 @@ impl Analyzer<'_, '_> {
...
-                        right: box r.clone(),
-                        cause: vec![],
+                return Ok(());
+            }
+            Type::Alias(alias_ty) => {
+                return self.assign_to_intrinsic(&mut Default::default(), to, &alias_ty.ty, Default::default());
+            }
+            Type::Ref(ref_ty) => {
+                if let RTsEntityName::Ident(id) = &ref_ty.type_name {
+                    if let Ok(Some(ItemRef::Owned(v))) = &self.find_type(&id.into()) {
```
kdy1, line 3097: `ItemRef` should never be matched with a pattern

kdy1: You can use `cloned().collect()` instead

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2883,61 +2931,208 @@ impl Analyzer<'_, '_> {
...
+                return Ok(());
+            }
+            Type::Param(param) => {
+                if let Some(constraint) = &param.constraint {
+                    if constraint.is_union_type() || constraint.is_type_param() {
+                        return self.assign_to_intrinsic(&mut Default::default(), to, constraint, Default::default());
+                    }
+
+                    if !constraint.is_str_like() && !constraint.is_never() {
+                        let mut span: Span = to.type_args.span();
```
kdy1, line 3112: `.or_else()` would be better

kdy1 (changes requested):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
awareness481: Just did, took the liberty of adding `AssignOpts` to `expand_intrinsic_types` as a param so I can pass `opts.span`

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: That makes sense but I don't think I understand where the `cloned.collect()` suggestion is supposed to fit in here :fearful:

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: We could do something like
```rs
                if type_param.is_alias() || type_param.is_ref_type() {
                    return self.assign(to.span(), &mut Default::default(), type_param, r);
                }
```

or were you thinking of calling `self.normalize` ?

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
kdy1: The latter, but I'm not 100% sure

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: removed the if clause completely

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: Removed this as well, it is never matched when running `./scripts/check.sh` and this way it won't be a false positive if we actually need to handle the `ref` match arm.

I believe I had added this when I was trying to make type alias validation work.

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -220,8 +221,21 @@ impl Analyzer<'_, '_> {
...
+                            ..self.ctx
+                        };
+
                         let ty = self
-                            .expand_intrinsic_types(actual_span, i)
+                            .with_ctx(ctx)
+                            .expand_intrinsic_types(
+                                actual_span,
+                                i,
+                                AssignOpts {
```
kdy1, line 234: I think we should pass `span`, not `AssignOpts`

kdy1: Logically the function for expanding accepting options are assignment is strange

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -1975,14 +1991,24 @@ impl Analyzer<'_, '_> {
         v
     }
 
-    pub(crate) fn expand_intrinsic_types(&mut self, span: Span, ty: &StringMapping) -> VResult<Type> {
+    pub(crate) fn expand_intrinsic_types(&mut self, span: Span, ty: &StringMapping, opts: AssignOpts) -> VResult<Type> {
```
kdy1, line 1994: `span_for_validation: Span` would be better

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2883,61 +2922,195 @@ impl Analyzer<'_, '_> {
...
 
-                    return Err(ErrorKind::AssignFailed {
-                        span: r.span(),
-                        left: box Type::StringMapping(to.clone()),
-                        right_ident: None,
-                        right: box r.clone(),
-                        cause: vec![],
+                return Ok(());
+            }
+            Type::Alias(alias_ty) => {
```
kdy1, line 3083: I think this branch should be `unreachable!()`

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2883,61 +2922,195 @@ impl Analyzer<'_, '_> {
...
+                return self.assign_to_intrinsic(&mut Default::default(), to, &alias_ty.ty, Default::default());
+            }
+            Type::Param(param) => {
+                if let Some(constraint) = &param.constraint {
+                    if constraint.is_union_type() || constraint.is_type_param() {
+                        return self.assign_to_intrinsic(&mut Default::default(), to, constraint, Default::default());
```
kdy1, line 3089: Please preserve `AssignData`

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -2883,61 +2922,195 @@ impl Analyzer<'_, '_> {
...
-                    return Err(ErrorKind::AssignFailed {
-                        span: r.span(),
-                        left: box Type::StringMapping(to.clone()),
-                        right_ident: None,
-                        right: box r.clone(),
-                        cause: vec![],
+                return Ok(());
+            }
+            Type::Alias(alias_ty) => {
+                return self.assign_to_intrinsic(&mut Default::default(), to, &alias_ty.ty, Default::default());
```
kdy1, line 3084: Please preserve `AssignData`

kdy1 (changes requested):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: Should I also preserve opts here?

awareness481: Done but also preserved opts

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
awareness481: You're right, I had assumed it was for this case
```ts
type B = Lowercase<string>;
type C = Lowercase<B>;
```
but here B is `ref type` which I means I need to re-add the `type::ref` match arm :skull:

awareness481: Actually, we need to it because the ref type will be converted to an alias type. What do you think about this?

```rs
            Type::Alias(alias_ty) => {
                return self.assign_to_intrinsic(data, to, &alias_ty.ty, opts);
            }
            Type::Ref(ref_ty) => {
                if let RTsEntityName::Ident(r_ident) = &ref_ty.type_name {
                    if let Ok(Some(ty)) = self.find_type(&r_ident.into()) {
                        let ty = ty.into_iter().map(|v| v.into_owned()).collect::<Vec<Type>>();
                        return self.assign_to_intrinsic(data, to, &ty.clone()[0], opts);
                    }
                }
            }
```

if we keep alias unreachable the below fails
```ts
type B = Lowercase<string>;
type C = Lowercase<B>;
```

`B` is first `ref` then gets converted to `alias` in in `self.find_type`

awareness481: Removed `Type::Alias` from this match and added a `Type::Ref` match arm in `expand_intrinsic_types` function.

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -2109,21 +2130,94 @@ impl Analyzer<'_, '_> {
...
+        #[allow(clippy::question_mark)]
+        if let Ok(ref ty) = normalized_ty {
+            if let Type::StringMapping(ty) = ty.normalize() {
+                match ty.type_args.params[0].normalize() {
+                    Type::Ref(ref_ty) => {
+                        if let RTsEntityName::Ident(r_ident) = &ref_ty.type_name {
+                            if let Ok(Some(ty_found)) = self.find_type(&r_ident.into()) {
+                                let ty_found = &ty_found.into_iter().map(|v| v.into_owned()).collect::<Vec<Type>>()[0];
+
+                                if let Type::Alias(alias) = ty_found.normalize() {
```
kdy1, line 2156: Is this logic really required?

kdy1: I'm considering removing `Type::Alias` completely, as it's quite useless type because it requires proper normalization anyway.

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -2109,21 +2130,94 @@ impl Analyzer<'_, '_> {
...
+                                            },
+                                            metadata: ty.metadata,
+                                        },
+                                        span_for_validation,
+                                    );
+                                }
+                            }
+                        }
+                    }
+                    Type::Alias(alias) => {
```
kdy1, line 2188: This is wrong

kdy1: This should be unreachable

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
kdy1: This variant should be processed by `expand()`, `expand_top_ref` or `normalize()`

crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
awareness481: With the assumption that `expand_top_ref` never returns alias (so it always returns the type of alias) this not needed. I removed type::alias logic and the following still works correctly

```ts
type A = number;
type B = A;
type C = Lowercase<B>; // should error here
```
so I think its ok

kdy1 (approved):
Thank you so much!

PR #801, opened 2023-03-06
Feature: last overflow from typeof this

sunrabbit123:
**Description:**
```ts
function Test3(this: { no: number }) {
  let x: typeof this.no = 1;
}
```


crates/stc_ts_file_analyzer/src/analyzer/stmt/var_decl.rs
```diff
@@ -262,33 +262,68 @@ impl Analyzer<'_, '_> {
...
+                                ty.normalize(),
                                 Type::Query(QueryType {
-                                    span,
-                                    expr: box QueryExpr::TsEntityName(RTsEntityName::Ident(RIdent::new(
-                                        "globalThis".into(),
-                                        span.with_ctxt(SyntaxContext::empty()),
-                                    ))),
-                                    metadata: Default::default(),
-                                    tracker: Default::default(),
+                                    expr: box QueryExpr::TsEntityName(RTsEntityName::TsQualifiedName(box RTsQualifiedName {
```
kdy1, line 269: I think this branch is reachable using other types than `globalThis`

kdy1: Please use `ty.is_global_this()` instead.
https://rustdoc.stc.dudy.dev/stc_ts_types/enum.Type.html#method.is_global_this

kdy1 (changes requested):
Thank you!

crates/stc_ts_file_analyzer/src/analyzer/stmt/var_decl.rs
sunrabbit123: this code mean `typeof this`

sunrabbit123: just contains globalThis, this mean never only globalThis

sunrabbit123: Do symbols also use query types?

sunrabbit123: I'll check it out tomorrow.

crates/stc_ts_file_analyzer/src/analyzer/stmt/var_decl.rs
```diff
@@ -262,33 +262,68 @@ impl Analyzer<'_, '_> {
...
+                                        left: RTsEntityName::Ident(..),
```
kdy1, line 270: `Ident(..)` => `Ident(RIdent { sym: js_word!("this"), .. })`

kdy1 (approved):
Thanks!

kdy1 (changes requested):
Can you rebase? I tried rebasing but seems like there's a conflict

kdy1 (approved):
Thanks!

PR #803, opened 2023-03-07
fix: fix #795

### LINKED ISSUE #795: If the union type has `any`,  it should  infer to be `any`
In this case:
```ts
declare const a: string | any;
a;
```
tsc infers the type of `a` to be `any`.
However, the stc infers it as
```
 1 | declare const a: string | any;
 2 | a;
   : ^
   `----

Error: 
  > (string | any)
```

In my opinion,   The type of `a` should be as tsc infers.

blacktoast:
**Description:**
Fixed based on comments in #795

If `any` is in the union type, `any` is inferred.

**BREAKING CHANGE:**

#795

### DIFF
--- a/crates/stc_ts_file_analyzer/tests/pass/union/any/1.swc-stderr
+++ b/crates/stc_ts_file_analyzer/tests/pass/union/any/1.swc-stderr
@@ -1,0 +1,41 @@
+
+  x Type
+   ,-[$DIR/tests/pass/union/any/1.ts:7:1]
+ 7 | const t1: Any = 1;
+ 8 | t1;
+   : ^^
+   `----
+
+Error: 
+  > any
+
+  x Type
+    ,-[$DIR/tests/pass/union/any/1.ts:9:1]
+  9 | 
+ 10 | const t2 = a;
+    :            ^
+ 11 | t2;
+    `----
+
+Error: 
+  > any
+
+  x Type
+    ,-[$DIR/tests/pass/union/any/1.ts:10:1]
+ 10 | const t2 = a;
+ 11 | t2;
+    : ^^
+    `----
+
+Error: 
+  > any
+
+  x Type
+    ,-[$DIR/tests/pass/union/any/1.ts:13:1]
+ 13 | const t3: Any2 = 1;
+ 14 | t3;
+    : ^^
+    `----
+
+Error: 
+  > any
--- a/crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
@@ -341,7 +341,12 @@ impl Analyzer<'_, '_> {
                                     .normalize(span, Cow::Borrowed(ty), opts)
                                     .context("tried to normalize an element of a union type")?;
                                 ty.freeze();
+
                                 let mut ty = ty.into_owned();
+
+                                if ty.is_any() && !opts.process_only_key {
+                                    return Ok(Cow::Owned(ty));
+                                }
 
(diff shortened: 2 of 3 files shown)


crates/stc_ts_file_analyzer/tests/pass/types/keyof/3.swc-stderr
```diff
@@ -8,7 +8,7 @@
     `----
 
 Error: 
-  > ("a" | "c" | string | number | symbol)
```
kdy1, line 11: This should not be changed

crates/stc_ts_file_analyzer/tests/pass/types/keyof/3.swc-stderr
```diff
@@ -19,7 +19,7 @@ Error:
...
```
kdy1, line 22: Same here

kdy1 (changes requested):

crates/stc_ts_file_analyzer/tests/pass/types/keyof/3.swc-stderr
blacktoast: in tsc ([link](https://www.typescriptlang.org/ko/play?#code/PTACGcBcCcEsGNIC4AEMCuBTAUCFoATTeAGwENozJYB7AO1Qx21LPHBQEEUBvbFASjKoocOgHN+g+KjroAtgCNM0bAF9s2TAA8ADjWiQ0AT12YUAaUzGOAXhQBrazQBmKABTcAPkLrGAlJpErNDmAG4UjtbgqFY2QcTkoSgkmEYuNDQiMLASKD5ySir5KODGSjQkmhk0KPZO8Q12KDVAA)), `foo`, `keys` are both inferred as (string | number | symbol). 

I don't understand why it should be inferred as ("a" | "c" | string | number | symbol)

Is there something I'm missing or need to know?

kdy1: Those strings are important for a language server

kdy1: The information shown on hover is not a complete information.

blacktoast:
Changed to not return `any` type if `keyof` is used.

crates/stc_ts_file_analyzer/tests/pass/union/any/1.swc-stderr
```diff
@@ -0,0 +1,41 @@
+
+  x Type
+   ,-[$DIR/tests/pass/union/any/1.ts:7:1]
+ 7 | const t1: Any = 1;
+ 8 | t1;
+   : ^^
+   `----
+
+Error: 
+  > any
```
kdy1, line 10: This should not be any.

https://www.typescriptlang.org/play?#code/C4TwDgpgBAKlC8UCGA7EUA+UDOwBOAligOaZQDkAZgPbXlnkBGSe5A3FAFDeeUCuKAMbAC1FFD7YIAQTQAKYAC5YASigBvTlG1QClKAoTxEVcis4BfHkA
image

kdy1 (changes requested):

crates/stc_ts_file_analyzer/tests/pass/union/any/1.swc-stderr
kdy1: I think the problem here is that https://github.com/dudykr/stc/issues/795 is a wrong issue from start

kdy1: We cannot drop other types even if `any` is a member of a union

PR #812, opened 2023-03-12
fix(ts/analyzer): improve type-guard of 'in' operator

togami2864:
**Description:**
If `i` were an index signature, false facts should not be created unlike a union type. This caused the checker to be unable to access the type of `i` in the else clause.
```
interface Indexed {
  [s: string]: any;
}

function f(i: Indexed) {
  if ("a" in i) {
    return i.a;
  } 
  return "c" in i && i.c; // error: i is inferred as never
}
```

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #813, opened 2023-03-12
fix #787

### LINKED ISSUE #787: Allow more return types for async generators
Test case

https://github.com/dudykr/stc/blob/19aab18c6b8e1d71373e3345a0c335f58596bddd/crates/stc_ts_type_checker/tests/conformance/types/asyncGenerators/types.asyncGenerators.es2018.1.ts#L118-L120

We should modify the logic to avoid assigning to the declared return type if a yield expression is in an async generator.

We can do it by checking if it's async iterable before assigning it from

https://github.com/dudykr/stc/blob/19aab18c6b8e1d71373e3345a0c335f58596bddd/crates/stc_ts_file_analyzer/src/analyzer/stmt/return_type.rs#L498-L511

blacktoast:
**Description:**
It's not done yet, but I've opened a PR to ask if this is the correct direction I'm headed.

I fixed it to not do assign when it's an async generator.
So the` types.asyncGenerators.es2018.1.ts `test passed, but the `types.asyncGenerators.es2018.2.ts` test result changed.

My guess is that in the types.asyncGenerators.es2018.2.ts test, a case like 
```ts
async function * explicitReturnType4(): AsyncIterable<number> {
    yield "a";
}
```
returned the correct error, in 
https://github.com/dudykr/stc/blob/19aab18c6b8e1d71373e3345a0c335f58596bddd/crates/stc_ts_file_analyzer/src/analyzer/stmt/return_type.rs#L515  returned an error even though it was the correct type.

So I think I need to add some other logic to return the correct error on `in_async`, am I right or is my approach wrong?

**BREAKING CHANGE:**


kdy1:
It's the wrong direction. We should still assign it even if it's an async generator.
Our focus is not matching the numbers - instead, we should focus on making inference/rule correct

blacktoast:
thanks, I'll try again.

PR #816, opened 2023-03-14
Handle intrinsic case where both args are generic

awareness481:
**Description:**

```ts
function foo<T extends "HELLO", U extends T>(x: Uppercase<T>, u: Uppercase<U>) {
  u = x; // bad U (lhs) extends T (rhs)
  x = u; // good T (lhs) does not extend U (rhs)
}
```


awareness481:
Just found that this

```ts
function foo<T extends string, U extends "HELLO">(
  x: Uppercase<T>,
  u: Uppercase<U>
) {
  u = x; // should error
  x = u; // should error
}
```
only reports 1 error so im moving this back to draft, sorry!

kdy1 (approved):
Thanks!

PR #825, opened 2023-03-14
fix pass/stringMappingOverPatternLiterals/1.ts

awareness481:
**Description:**

I have no idea why this works so merge at your own risk


kdy1 (changes requested):
This PR is wrong. If you enable logging, there's message like

```
  ERROR unimplemented: assign:
Left: number;
Right: `${number}`;
```

kdy1:
I added `--nocapture` to `base.sh` of the file analyzer.

```diff
-UPDATE=1 cargo test -p stc_ts_file_analyzer $@ --color always --lib --test base --features no-threading -- -Zunstable-options --report-time
+UPDATE=1 cargo test -p stc_ts_file_analyzer $@ --color always --lib --test base --features no-threading -- -Zunstable-options --report-time --nocapture
```

crates/stc_ts_file_analyzer/src/analyzer/assign/tpl.rs
```diff
@@ -74,6 +82,49 @@ impl Analyzer<'_, '_> {
...
 
+                if target.is_bool() {
+                    let value = &*value.value;
+
+                    match value {
+                        "true" | "false" => {
+                            return Ok(true);
+                        }
+                        "TRUE" | "FALSE" if !data.dejavu.is_empty() => {
+                            if let Type::Instance(ty) = &data.dejavu[0].0.normalize() {
```
awareness481, line 93: No idea if relying on `data.dejavu` here is problematic or not but I wanted to consult you before trying anything else since I think other solutions will be significantly harder

kdy1: You should not use `data.dejavu`

awareness481:
No point in keeping this open as it needs significant changes to work

PR #832, opened 2023-03-16
feat: Handle assignment of indexed access types

kdy1:
**Description:**

 - Handle assignments of indexed access types.


kdy1:
I'll fix regressions with follow-up PRs.

PR #833, opened 2023-03-16
feat: Implement more logic for classes

kdy1:
**Description:**

 - Validate index signatures of a class first.
 - Handle `this` in `access_property`.

---

I extracted this from https://github.com/dudykr/stc/pull/832


kdy1:
Closing as this does not change test stats

PR #835, opened 2023-03-17
fix: Be strict about converting `Enum` to `EnumVariant`

kdy1:
**Description:**


kdy1:
I found that this is wrong

PR #838, opened 2023-03-19
fix: Fix `keyof` of type parameters

kdy1:
**Description:**

We should infer it as a `keyof T` instead of `string`.


kdy1:
Closing as this was wrong

PR #843, opened 2023-03-21
fix: Assign the object of indexed access type to type elements

togami2864:
**Description:**

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

PR #967, opened 2023-04-02
fix: instantiating non-generic with type args 

awareness481:
**Description:**
```ts
var f: { (): void };
export var r2 = new f<number>(); // 2558 instead of 2350
```

closes https://github.com/dudykr/stc/issues/946


kdy1 (approved):
Thank you!

PR #964, opened 2023-04-01
feat: Improve `castable`

kdy1:
**Description:**

**Related issue (if exists):**

Issue #921, opened 2023-03-25
Fix unit test:  `tests/pass-only/conformance/types/conditional/conditionalTypes1/.9.ts`


kdy1:
Related test input: https://github.com/dudykr/stc/blob/main/crates/stc_ts_file_analyzer/tests/pass-only/conformance/types/conditional/conditionalTypes1/.9.ts

Test file name: `tests/pass-only/conformance/types/conditional/conditionalTypes1/.9.ts`

---

I (`@kdy1`) may expand this in the future.
If you want to work on this issue, please ping `@kdy1` in the comment.

---

This issue is created by sync script.

kdy1:
Fixed by https://github.com/dudykr/stc/pull/964

PR #968, opened 2023-04-04
Feature(analyzer): fix type casting with conditional types

sunrabbit123:
**BREAKING CHANGE:**
```ts
type ZeroOf<T extends number | string | boolean> = T extends number ? 0 : T extends string ? "" : false;

function zeroOf<T extends number | string | boolean>(value: T) {
    return <ZeroOf<T>>(typeof value === "number" ? 0 : typeof value === "string" ? "" : false);
}

export function f21<T extends number | string>(x: T, y: ZeroOf<T>) {
    let z1: number | string = y;
    let z2: 0 | "" = y;
}
```


crates/stc_ts_file_analyzer/src/analyzer/expr/type_cast.rs
```diff
@@ -106,7 +106,7 @@ impl Analyzer<'_, '_> {
         casted_ty.freeze();
 
         self.validate_type_cast_inner(span, &orig_ty, &casted_ty).report(&mut self.storage);
-
+        dbg!(&casted_ty);
```
kdy1, line 109: Please remove this

kdy1 (changes requested):

kdy1 (approved):
Thanks!

sunrabbit123:
this resolve #922 #921

kdy1:
Did you try `./auto-unignore.sh`? If CI is not broken, CI should fail if it's really resolved

sunrabbit123:
this duplicate pr #964
The work overlaps, is it okay?

It looks like the difference between front-end and back-end processing.
And the code on the #964 looks cleaner and has less overhead

kdy1:
Ah, I'll revert this then

kdy1:
Wait, do you think it's fine to revert this PR? Do two PRs semantically identical?

sunrabbit123:
#964 recursively processes a conditional value.
#968 recursively pulls out the true and false values from the front of the conditional.

The difference is that #968 extracts the union type and uses the
#964 recursively checks them one by one.

I think they mean the same thing, but

I'm curious to know what your opinion is.

Issue #948, opened 2023-03-25
Fix unit test:  `tests/tsc/conformance/types/union/unionTypeConstructSignatures/.1.ts`


kdy1:
Related test input: https://github.com/dudykr/stc/blob/main/crates/stc_ts_file_analyzer/tests/tsc/conformance/types/union/unionTypeConstructSignatures/.1.ts

Test file name: `tests/tsc/conformance/types/union/unionTypeConstructSignatures/.1.ts`

---

I (`@kdy1`) may expand this in the future.
If you want to work on this issue, please ping `@kdy1` in the comment.

---

This issue is created by sync script.

sunrabbit123:
It's normal to get an error when you run this.
Is this a case where the argument min must be exactly 2 in the error history to pass?

kdy1:
No, the error should match the one from `tsc`

sunrabbit123:
> No, the error should match the one from tsc

```rs
ErrorKind::ExpectedNArgsButGotM { .. } => 2554, 
ErrorKind::ExpectedAtLeastNArgsButGotM { .. } => 2555,
```

The error was different!
I was confused by the name :D

thx.

PR #980, opened 2023-04-11
feat: Improve logic for selecting overload

sunrabbit123:
**Description:**
```ts
var strOrNum: string | number;

export var unionWithRestParameter2:
  | { (a: string, ...b: number[]): string }
  | { (a: string, b: number): number };
strOrNum = unionWithRestParameter2(); // error no call signature
```

**Related issue:**

 - Closes https://github.com/dudykr/stc/issues/948.
 - Closes https://github.com/dudykr/stc/issues/949.


crates/stc_ts_type_checker/tests/conformance/types/union/unionTypeConstructSignatures.error-diff.json
```diff
@@ -14,19 +15,21 @@
...
   },
   "extra_errors": {
-    "TS2555": 1
+    "TS2554": 1
   },
   "extra_error_lines": {
-    "TS2555": [
-      63
+    "TS2554": [
+      70
```
sunrabbit123, line 32: should fix it

sunrabbit123:
If the args are empty and all WrongArgCount, then the
treat the rest as 1 and the normal args as 10, summing them up and returning a 
and picking the largest value seems like the correct reasoning.
Of course, we're talking about duplicates after a traditional sort.

Issue #956, opened 2023-03-25
Improve Interface type params


awareness481:
- [ ] Interface type params not validated
   ```ts
   interface A<T extends string> {}
   type a = A<number> // required error, not reported
   ```
- [x] If two interfaces of the same name and the second one has more type params, there's a panic.
  ```ts
  interface A<T extends string> {}
  interface A<T extends number, U extends number> {}
  type a = A<string, number>; // panics
  ```
  ```rs
  thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 1', 
  crates/stc_ts_file_analyzer/src/analyzer/decl_merging.rs:109:43
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  ```

awareness481:
If you agree with the issue, please assign me as I want to try fixing this, thanks ^^

kdy1:
Thank you!

sunrabbit123:
@awareness481 
Are you running `Interface type params not validated`?
I'm getting an error on that part, so I'm guessing you need to assign a worker to it.

The problem is caused by propagating generics and not checking for type.

I think the right place to work on it is in analyzer/generic/mod.rs:122.

If you're working on it, you can use the
How far along are you?

awareness481:
@sunrabbit123 I was actually thinking of removing the assignment from myself because when I tried it I couldn't get it working. Do you want to give it a go yourself? Please feel free to

sunrabbit123:
I've been playing around with it a bit now, and I've found that
Unfortunately, it's working fine, but I'm getting a lot of regressions. 😭

sunrabbit123:
revert this issue
#1029

sunrabbit123:
#1053 

Close the issue because it has been implemented in that location.

PR #985, opened 2023-04-14
handle different arity of type params for interfaces

awareness481:
**Description:**

Ref: https://github.com/dudykr/stc/issues/956


crates/stc_ts_file_analyzer/src/analyzer/decl_merging.rs
```diff
@@ -97,24 +97,37 @@ impl Analyzer<'_, '_> {
...
-                // TODO: Handle the number of type parameters.
                 let mut type_params = FxHashMap::default();
-                if let Some(b_tps) = &bi.type_params {
-                    if let Some(a_tp) = &a.type_params {
+
+                match (&a.type_params, &bi.type_params) {
+                    (Some(a_tps), Some(b_tps)) => {
+                        if a_tps.params.len() != b_tps.params.len() {
+                            self.storage
+                                .report(ErrorKind::InterfaceNonIdenticalTypeParams { span: a.span() }.into());
```
awareness481, line 106: span currently looks like this
```rs
────────────────────────────────
```
is there any way to only use the span of the name of the interface? (like tsc)

```rs
```rs
          ─
```
```

File: crates/stc_ts_file_analyzer/src/analyzer/decl_merging.rs [hunk 2]
```diff
@@ -97,24 +97,37 @@ impl Analyzer<'_, '_> {
...
+                        }
+
                         for (idx, b_tp) in b_tps.params.iter().enumerate() {
                             type_params.insert(
                                 b_tp.name.clone(),
                                 Type::Param(TypeParam {
-                                    span: a_tp.span,
-                                    name: a_tp.params[idx].name.clone(),
+                                    span: a_tps.span,
+                                    name: a_tps.params.get(idx).unwrap_or(b_tp).name.clone(),
```
awareness481, line 116: Does it matter here that `a_tps.span` can be passed when `a_tps.params.get(idx) = None` ?

crates/stc_ts_file_analyzer/src/analyzer/decl_merging.rs
kdy1: I think we should add a field like `name_span` to the `Interface` type

crates/stc_ts_file_analyzer/src/analyzer/decl_merging.rs
kdy1: Hmm, not sure, but I don't think it matters.

kdy1 (approved):
Thank you!

PR #987, opened 2023-04-15
handle different names for type params

awareness481:
**Description:**

```ts
interface A<T> {}
interface A<U> {}

// both should error because of different names for type params
```

see: https://github.com/dudykr/stc/blob/main/crates/stc_ts_type_checker/tests/conformance/interfaces/declarationMerging/twoGenericInterfacesDifferingByTypeParameterName.ts


crates/stc_ts_file_analyzer/src/analyzer/decl_merging.rs
```diff
@@ -109,6 +111,15 @@ impl Analyzer<'_, '_> {
                         }
 
                         for (idx, b_tp) in b_tps.params.iter().enumerate() {
+                            if let Some(a_tp) = a_tps.params.get(idx) {
+                                if !is_builtin_type && a_tp.name.sym() != b_tp.name.sym() {
```
awareness481, line 115: Context: If the user defined a builtin interface:

```ts
interface Array<T> {...}
```

then we don't (and should not) check the type param name. **However**, if the user defined multiple interfaces of a builtin interface

```ts
interface Array<T> {}
interface Array<X> {}
```
then we should check the type param names, and in this case return a `2488` error for both. Currently this PR doesn't handle the second case (with multiple builtin interfaces)

awareness481: The second case is pretty tricky because the declaration merge always happens against the builtin-type

kdy1: I'm sorry, but can you write it as comments? I'm sure I'll forget this after a few months of further development...

awareness481: Yes ofc!

awareness481: Added the comment, i'm not sure what's the preferred formatting or if you want to add anything, so please either add/change something directly or request changes to the PR.

kdy1 (approved):
Thank you!
I'll rebase/merge this soon

kdy1 (approved):

PR #990, opened 2023-04-15
Consider scope for type alias duplicate name error

awareness481:
**Description:**

```ts
// before
type A = any; // 2300 Duplicate name
{
  type A = any; // 2300 Duplicate name
  {
    type A = any; // 2300 Duplicate name
  }
}

// after

type A = any; // no error
{
  type A = any; // no error
  {
    type A = any; // no error
  }
}
```

Searched the relevant tests in the TypeScript repo but I couldn't find specific tests that test type alias with scope in mind. Nevertheless this seems to be the correct behavior (no error if different block scope)


kdy1 (changes requested):
Can you add some tests to https://github.com/dudykr/stc/tree/8a582c7a3f23d0d0b54a8e3b58fcc511cdb41e30/crates/stc_ts_file_analyzer/tests/tsc ?

kdy1 (approved):
Thank you!

PR #1003, opened 2023-04-20
handle boolean inside template literal as union

awareness481:
**Description:**


crates/stc_ts_file_analyzer/tests/pass/types/literal/stringMappingOverPatternLiterals/2.swc-stderr
```diff
@@ -19,4 +19,4 @@ Error:
    `----
 
 Error: 
-  > `AA${number}${bigint}${boolean}`
+  > `AA${number}${bigint}${(TRUE | FALSE)}`
```
awareness481, line 22: So these are generated by running the rust runner I presume but I dont think I understand the output. Is there any way I can confirm whether this is correct or not? :thinking:

kdy1: You can copy the input typescript file to the typescript playground and hover on the type to see the _correct_ type

awareness481: Great, thanks!

awareness481: Initially I just wanted to fix the one error but I should try to do it the right way (similarly how tsc does it) so it has the correct types as well and covers more cases. so I closed it for now

kdy1: I see, thank you!

PR #1004, opened 2023-04-20
fix typo in `build` script error message

DetachHead:


kdy1:
I'm sorry, but we don't accept PRs like this

PR #1011, opened 2023-05-03
Feature(analyzer): Implement typeparam validation

### LINKED ISSUE #956: Improve Interface type params

sunrabbit123:
**Description:**
```ts
type a = A<number> // required error, not reported
```

**Related issue:**

 - Closes #956.


sunrabbit123:
We're seeing a significant amount of regression, which is something we need to fix.

sunrabbit123:
At the first
```ts
{
    get x() {
      return "boolean" as const;
    },
    y: "string",
  }
```
The above values should be deduced as follows
```ts
{
    readonly x: "boolean";
    y: "string";
}
```
And in `stc`, it's reasoned like this
```
  > {
  |     readonly x: any;
  |     y: "string";
  | }
```

sunrabbit123:
```ts
type A = string[] extends unknown[] ? string : never;
//   ^?
type B<T extends unknown[]> = T;
type C = B<string[]>;
```
is not working

sunrabbit123:
[crates/stc_ts_type_checker/tests/conformance/types/tuple/variadicTuples1](https://github.com/dudykr/stc/pull/1011/files#diff-65ddc58aa56c3d8be495ff4e2bbf6338ed82ca8a5ef48af8e18af77ac27c1c23)

That test has different results in file_analyzer and type_checker.

kdy1 (approved):
Thank you!

kdy1:
I partially revert this PR with https://github.com/dudykr/stc/pull/1029 because the regression of performance was way too much

sunrabbit123:
In my opinion, the following features are required

https://github.com/dudykr/stc/pull/1029/files#diff-35cb11bec0b1ce7291514626e7a22ef1b38573bd88263c9226163ebb533f79eeL571-L583

kdy1:
Can you retry this PR, but step-by-step this time?

sunrabbit123:
We'll work on them one by one at a later time

PR #1020, opened 2023-05-11
feat(ts/analyzer): support `noUncheckedIndexedAccess`

togami2864:
**Description:**

**BREAKING CHANGE:**


crates/stc_ts_env/src/lib.rs
```diff
@@ -222,6 +223,7 @@ impl From<&CompilerOptions> for Rule {
             allow_unused_labels: v.allow_unused_labels.unwrap_or_default(),
             no_fallthrough_cases_in_switch: v.no_fallthrough_cases_in_switch.unwrap_or_default(),
             no_implicit_returns: v.no_implicit_returns.unwrap_or_default(),
+            no_unchecked_indexed_access: false,
```
togami2864, line 226: i can add this feature to [tsconfig](https://github.com/drivasperez/tsconfig) directory(I'm not sure the maintainer will response). Do you keep using this crate??

crates/stc_ts_type_checker/tests/conformance/pedantic/noUncheckedIndexedAccess.stats.rust-debug
```diff
@@ -1,6 +1,6 @@
 Stats {
-    required_error: 24,
-    matched_error: 7,
-    extra_error: 0,
+    required_error: 5,
+    matched_error: 26,
+    extra_error: 1,
```
togami2864, line 4: I'll report it as an issue since it's a bit complicated and just a minor bug.

crates/stc_ts_env/src/lib.rs
kdy1: I think vendoring it would worth the effort.

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -4355,6 +4355,47 @@ impl Analyzer<'_, '_> {
...
             Ok(ty.union_with_undefined(span))
         } else {
+            if self.rule().no_unchecked_indexed_access && matches!(self.ctx.pat_mode, PatMode::Decl) {
+                let field_is_symbol = match &prop {
+                    Key::Computed(key) => key.ty.is_symbol_like(),
+                    _ => false,
+                };
+
+                let is_variadic_tuple = match &obj_ty {
+                    Type::Arc(_) => match obj_ty.normalize() {
```
kdy1, line 4365: I think this can be problematic in the future, as `Type::Arc` should be used in more cases and it's not for _variadic_. That's why I always use `Type::normalize` while pattern matching

kdy1 (approved):
Thank you!

PR #1023, opened 2023-05-15
feat: Improve `extends` of `readonly` types

sunrabbit123:
**Description:**

```ts
type A = readonly string[] extends unknown[] ? string : never;

let a: A = "1" as never;
a; // never;

type B = string[] extends readonly unknown[] ? string : never;
let b: B = "1" as string;
b; // string

type C = readonly string[] extends readonly unknown[] ? string : never;
let c: C = "1" as string;
c; // string
```

#1011


kdy1 (changes requested):
Can you add changes of test stats files?

sunrabbit123:
As far as I can tell, there were no changes to the stats.

crates/stc_ts_type_checker/tests/conformance.pass.txt
```diff
@@ -1414,6 +1414,8 @@ expressions/binaryOperators/arithmeticOperator/arithmeticOperatorWithOnlyNullVal
 expressions/binaryOperators/arithmeticOperator/arithmeticOperatorWithTypeParameter.ts
 expressions/binaryOperators/arithmeticOperator/arithmeticOperatorWithUndefinedValueAndInvalidOperands.ts
 expressions/binaryOperators/arithmeticOperator/arithmeticOperatorWithUndefinedValueAndValidOperands.ts
+expressions/binaryOperators/comjsdoc/parseLinkTag.ts
```
kdy1, line 1417: Then why did you modify this file?

sunrabbit123: We know that location is invalid.

We suspect a bug was introduced while running the script

kdy1: Then please revert this file

sunrabbit123: I get it.
Do you know anything about this anomaly?

kdy1: You should not commit `pass.txt` by its own.

sunrabbit123: I was wondering why such a non-existent file would come out because I was committing the results of the script

kdy1 (approved):

PR #740, opened 2023-02-22
fix: Fix assignment of unions of intersections

kdy1:
**Description:**

**Related issue:**

 - Closes https://github.com/dudykr/stc/issues/415.

PR #1028, opened 2023-05-18
feat: Improve inference of a function argument with a literal type

sunrabbit123:
**Description:**

```ts
// @strict: true
// @target: esnext

declare function f1<T>(x: T, y: string | T): T;

export const a6 = f1(true, false);  // boolean
```

**Related issue:**

 - Closes #740.


kdy1 (approved):
Thank you!

PR #1031, opened 2023-05-19
Feat: fix recursive panic

sunrabbit123:
**Description:**

```ts
// The following are errors because of circular references
var c: typeof c;
var c: any;
var d: typeof e;
var d: any;
var e: typeof d;
var e: any;

interface Foo<T> {}
var f: Array<typeof f>;
var f: any;
var f2: Foo<typeof f2>;
var f2: any;
var f3: Foo<typeof f3>[];
var f3: any;

// None of these declarations should have any errors!
// Truly recursive types
var g: { x: typeof g };
var g: typeof g.x;
var h: () => typeof h;
var h = h();
var i: (x: typeof i) => typeof x;
var i = i(i);
var j: <T extends typeof j>(x: T) => T;
var j = j(j);

// Same as h, i, j with construct signatures
var h2: new () => typeof h2;
var h2 = new h2();
var i2: new (x: typeof i2) => typeof x;
var i2 = new i2(i2);
var j2: new <T extends typeof j2>(x: T) => T;
var j2 = new j2(j2);

// Indexers
var k: { [n: number]: typeof k; [s: string]: typeof k };
var k = k[0];
var k = k[""];

// Hybrid - contains type literals as well as type arguments
// These two are recursive
var hy1: { x: typeof hy1 }[];
var hy1 = hy1[0].x;
var hy2: { x: Array<typeof hy2> };
var hy2 = hy2.x[0];

interface Foo2<T, U> {}

// This one should be an error because the first type argument is not contained inside a type literal
var hy3: Foo2<typeof hy3, { x: typeof hy3 }>;
var hy3: any;
```


crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -462,13 +462,6 @@ impl Analyzer<'_, '_> {
                                     )));
                                 }
 
-                                if expanded_ty.is_query() {
-                                    unreachable!(
```
kdy1, line 466: Why did you remove this? I'm a bit worried about a stack overflow

sunrabbit123:
```ts
var g: { x: typeof g };
var g: typeof g.x;
```
`g.x` is `typeof g`

Eventually, the infinite recursion will get any from normalize and terminate.

crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs
```diff
@@ -613,7 +613,7 @@ impl Analyzer<'_, '_> {
         };
 
         // It's valid to assign any to everything.
-        if rhs.is_any() {
+        if rhs.is_any() || to.is_any() {
```
kdy1, line 616: Seems redundant.
https://github.com/dudykr/stc/blob/d932c27b1d597261cb37b9992868c63e01674e81/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs#L837-L839

sunrabbit123: That code is too far to reach and the
In the meantime, it's called back to normalize, causing an infinite recursion.

I decide it's better to deal with the obvious first.

kdy1: It should not be here. It's far because of good reasons

sunrabbit123: The code below is the function that is called to `normalize` before `to.is_any()` is called.
Code that calls `self.normalize()`, such as `Conditional` or `Query` with a recursive structure, will overflow before reaching `to.is_any()`.

Of course, the reason why `rust overflows` here is because we already got `any` with the overflow, but the recursion still doesn't end.
```rs
    fn normalize_for_assign<'a>(&mut self, span: Span, ty: &'a Type, opts: AssignOpts) -> VResult<Cow<'a, Type>> {
        ty.assert_valid();

        let ty = ty.normalize();

        if let Type::Instance(Instance { ty, .. }) = ty {
            // Normalize further
            if ty.is_ref_type() {
                let normalized = self.normalize_for_assign(span, ty, opts)?;

                if normalized.is_keyword() {
                    return Ok(normalized);
                }
            }

            if ty.is_mapped() {
                let ty = self.normalize_for_assign(span, ty, opts)?;

                return Ok(ty);
            }
        }

        match ty {
            Type::EnumVariant(e @ EnumVariant { name: Some(..), .. }) => {
                if opts.ignore_enum_variant_name {
                    return Ok(Cow::Owned(Type::EnumVariant(EnumVariant { name: None, ..e.clone() })));
                }
            }
            Type::Conditional(..)
            | Type::IndexedAccessType(..)
            | Type::Alias(..)
            | Type::Instance(..)
            | Type::StringMapping(..)
            | Type::Enum(..)
            | Type::Import(..)
            | Type::Tuple(..)
            | Type::Union(..)
            | Type::Index(..)
            | Type::Tpl(..)
            | Type::Query(..) => {
                let ty = self
                    .normalize(
                        Some(span),
                        Cow::Borrowed(ty),
                        NormalizeTypeOpts {
                            merge_union_elements: true,
                            preserve_global_this: true,
                            in_type_or_type_param: true,
                            ..Default::default()
                        },
                    )?
                    .into_owned();

                return Ok(Cow::Owned(ty));
            }
            _ => {}
        }

        Ok(Cow::Borrowed(ty))
    }
```

sunrabbit123: this cause next code
```ts
interface Foo<T> {}
var f: Array<typeof f>;
var f: any;
```

It turns out it was edge case code.
Is there a good direction to end this recursion?

kdy1: Ah, I think this change is inevitable, but we should move this to https://github.com/dudykr/stc/blob/d932c27b1d597261cb37b9992868c63e01674e81/crates/stc_ts_file_analyzer/src/analyzer/assign/mod.rs#L631

sunrabbit123: any can be moved,
but I think the fundamental problem is that you're trying to assign to normalize and normalize to assign.

kdy1: Please take a look at the actual source code. It's not related to normalize

sunrabbit123: Well, it turns out that assign is an execution trigger.
I'll attach the logs.
```ts
interface Foo<T> {}
var f: Array<typeof f>;
var f: any;
```
```
  DEBUG  [types/fact] (typeof f[] | any); => (typeof f[] | any);
TypeFacts: None
    at crates/stc_ts_file_analyzer/src/ty/type_facts.rs:151
    in apply_type_facts_to_type
    in apply_type_facts with name: f#2
    in type_of_var
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in normalize with ty: (typeof f[] | any);
    in normalize with ty: typeof f;
    in normalize with ty: typeof f;
    in normalize with ty: typeof f[];
    in normalize with ty: typeof f[];
    in instantiate_for_normalization with ty: "typeof f[];"
    in normalize with ty: instanceof  typeof f[];
    in assign with lhs: "any;", rhs: "instanceof  typeof f[];"
    in Stmt with line_col: "(3:1-3:12)"


thread 'pass_tests__pass__custom__1_ts' has overflowed its stack
fatal runtime error: stack overflow
error: test failed, to rerun pass `--test base`

Caused by:
  process didn't exit successfully: `/home/sunrabbit/Desktop/stc/target/debug/deps/base-7325589fa1148cfd pass__custom__ --nocapture` (signal: 6, SIGABRT: process abort signal)
```

PR #1035, opened 2023-05-28
feat: Implement assignment rules for accessibility

kdy1:
**Description:**


kdy1:
Closing as this does not worth the effort at now

PR #1038, opened 2023-05-29
feat: Report more errors for wrong class inheritance

kdy1:
**Description:**


kdy1:
Seems like my approach was wrong. I should use assign logic for it, but I'm not sure how can I do it in current design

Issue #377, opened 2022-12-06
Use correct error code for readonly properties


kdy1:
https://github.com/dudykr/stc/blob/d50b685dd80aaa0ca4227551bb9cd26af0fd7af4/crates/stc_ts_type_checker/tests/conformance/types/intersection/intersectionTypeReadonly.ts#L17

This should fail with `TS2540`, but currently stc fails with 

```
TS2339

  x context: tried to access property of an object to calculate type of a member expression
  | context: tried to access property of an object (Base#6917;
  | Member as {
  |     readonly value: number;
  | };, id_ctx = Var)
  | Prop=Normal { span: Span { lo: BytePos(276), hi: BytePos(281), ctxt: #0 }, sym: Atom('value' type=inline) }
  | NoSuchProperty {
  |     span: Span {
  |         lo: BytePos(
  |             271,
  |         ),
  |         hi: BytePos(
  |             281,
  |         ),
  |         ctxt: #0,
  |     },
  |     obj: Some(
  |         Arc(
  |             Freezed {
  |                 ty: Interface(
  |                     Interface {
  |                         span: Span {
  |                             lo: BytePos(
  |                                 271,
  |                             ),
  |                             hi: BytePos(
  |                                 275,
  |                             ),
  |                             ctxt: #0,
  |                         },
  |                         name: Base#6917,
  |                         type_params: None,
  |                         extends: [],
  |                         body: [
  |                             Property(
  |                                 PropertySignature {
  |                                     span: Span {
  |                                         lo: BytePos(
  |                                             24,
  |                                         ),
  |                                         hi: BytePos(
  |                                             47,
  |                                         ),
  |                                         ctxt: #0,
  |                                     },
  |                                     accessibility: None,
  |                                     readonly: true,
  |                                     key: Normal {
  |                                         span: Span {
  |                                             lo: BytePos(
  |                                                 33,
  |                                             ),
  |                                             hi: BytePos(
  |                                                 38,
  |                                             ),
  |                                             ctxt: #0,
  |                                         },
  |                                         sym: Atom('value' type=inline),
  |                                     },
  |                                     optional: false,
  |                                     params: [],
  |                                     type_ann: Some(
  |                                         Keyword(
  |                                             KeywordType {
  |                                                 span: Span {
  |                                                     lo: BytePos(
  |                                                         40,
  |                                                     ),
  |                                                     hi: BytePos(
  |                                                         46,
  |                                                     ),
  |                                                     ctxt: #0,
[...]

PR #1043, opened 2023-06-03
Feature(analyzer): implement readonly error at intersection

sunrabbit123:
**Description:**

resolve intersection readonly

#377


kdy1 (approved):
Thank you!

PR #1044, opened 2023-06-03
refactor: Rename options of `access_property`

kdy1:
**Description:**

Extracted from https://github.com/dudykr/stc/pull/1042to make review easier.


kdy1:
This was wrong

PR #1045, opened 2023-06-03
refactor: Use `&self` instead of `&mut` where possible

kdy1:
**Description:**

This is required for #1042, which requires to recurse into self.


kdy1:
Status:

I did lots of work, but I'm stuck. I may need to split **all** call/new into `&` and `&mut`, because of the reevaluation logic. The reevaluation logic uses `ValidateWith::validate_with`, which takes `&mut`. I can't use `&` for `ValidateWith::validate_with` because of other constraints like interop with `Visit`.

PR #1047, opened 2023-06-11
fix: "false" | "true" when bool tpl

awareness481:
```ts
declare function nullishes(x: `${null | undefined}`): void;
nullishes("false"); // error should be reported
```
Adds error to above code. But if `boolean` is included in the union, no error

```ts
declare function nullishes(x: `${null | undefined | boolean}`): void;
nullishes("false"); // no error
```


kdy1 (approved):
Thank you!

PR #1049, opened 2023-06-17
fix: consider tpl with no types as string

awareness481:
**Description:**

The following all failed

```ts
function f1(a: `100`) {
  let x: `${number}`;
  x = `${a}`; // incorrect 2322
}
function f2(a: `true`) {
  let x: `${boolean}`;
  x = `${a}`; // incorrect 2322
}
function f3(a: `false`) {
  let x: `${boolean}`;
  x = `${a}`; // incorrect 2322
}
```

This happened because the case of a `tpl` having 0 types was not handled.

No test case covered this so I added new ones in `pass-only`


kdy1 (approved):
Thank you!

PR #1050, opened 2023-06-18
fix: numeric strings to tpl bigint type

awareness481:
**Description:**

A `bigint` type of a tpl accepted values like `100_000` or `10e3` but that should not happen as it should only accept numeric strings (without separators or scientific notation)

```ts
declare function bigints(x: `${bigint}`): void;
bigints("100_000"); // should 2345
```

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/tsc_helper.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/tsc_helper.rs
@@ -72,7 +72,13 @@ impl Analyzer<'_, '_> {
         } else if s.starts_with("0b") || s.starts_with("0B") {
             BigInt::parse_bytes(s[2..].as_bytes(), 2)
         } else {
-            s.parse::<BigInt>().ok()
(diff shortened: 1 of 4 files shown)


crates/stc_ts_file_analyzer/src/analyzer/tsc_helper.rs
```diff
@@ -72,7 +72,13 @@ impl Analyzer<'_, '_> {
         } else if s.starts_with("0b") || s.starts_with("0B") {
             BigInt::parse_bytes(s[2..].as_bytes(), 2)
         } else {
-            s.parse::<BigInt>().ok()
+            // BigInt strings only accepts numbers
+            // 1000n or 1_000n are both considered invalid
+            if s.parse::<i64>().is_ok() {
+                s.parse::<BigInt>().ok()
```
awareness481, line 78: I feel like this can be done better (perhaps convert an i64 to BigInt ?) but I couldnt get something working

kdy1 (approved):
Thank you!

PR #1052, opened 2023-06-30
fix stmt ordering

padorang684:
**Description:**
fix the ordering of statements when variables are erroneously declared multiple times

```javascript
var a = {}
var { } = a;

let b = 'a';
var {  [b]: soSo,  ...a } = a; // this errorneous re-declaration of var `a` hoists the line above the declaration of `b`

``` 

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

PR #1053, opened 2023-07-05
feat: Validate inferred generic types

sunrabbit123:
**Description:**
```ts
type a = A<number> // required error, not reported
```

**BREAKING CHANGE:**

#956


kdy1 (approved):
Thank you!

PR #1059, opened 2023-08-08
Feat: not implement keyof this

sunrabbit123:
**Description:**

implement `keyof this`
```ts
// @strict: true
// @declaration: true

type OldDiff<T extends keyof any, U extends keyof any> = (
    & { [P in T]: P; }
    & { [P in U]: never; }
    & { [x: string]: never; }
)[T];
interface A {
    a: 'a';
}
interface B1 extends A {
    b: 'b';
    c: OldDiff<keyof this, keyof A>;
}
type c1 = B1['c']; // 'c' | 'b'
```


crates/stc_ts_file_analyzer/src/analyzer/types/mod.rs
```diff
@@ -578,12 +578,13 @@ impl Analyzer<'_, '_> {
                                 return Ok(ty);
                             }
 
-                            let ty = self
-                                .normalize(span, Cow::Owned(prop_ty), opts)
-                                .context("tried to normalize the type of property")?
-                                .into_owned();
-
-                            return Ok(Cow::Owned(ty));
+                            // TODO(): scope of normalize should be obj_ty scope
```
sunrabbit123, line 581: Other than with_child, is there a better way to specify the scope as that object?

kdy1: You mean you only want to change `scope.this`?

kdy1: If so, I simply modifed `self.scope.this` like https://github.com/dudykr/stc/blob/81792e8e9dcfd95c4e3e0b98001a8ec47e0af27c/crates/stc_ts_file_analyzer/src/analyzer/class/mod.rs#L2461

kdy1: If you want to ensure that you don't forget resetting `self.scope.this`, you can do like

```rust

let old_this = self.scope.this.take();
self.scope.this = new_this;
let result = (|| {
    // your code.
})();
self.scope.this = old_this;

```

sunrabbit123: I want a scope outside of the traditional context.
Like when you call a closure, the closure doesn't use the caller's scope.

Of course, it's fine to do it now or with the code you suggested for now.

kdy1 (approved):
Thanks!

PR #1062, opened 2023-08-09
Feat: Implement this type in method parameter

sunrabbit123:
**Description:**

```ts
// @strictNullChecks: true
// @declaration: true


class A<T> {
	props: T & { foo: string };
}

class B extends A<{ x: number}> {
	f(p: this["props"]) {
		p.x;
	}
}
```

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

PR #1066, opened 2023-08-13
refactor: Create `Type::get_any_key_type`

sunrabbit123:
**Description:**
We've isolated duplicate types because it's unpleasant to look at the code if the keyword is stretched every time.

**BREAKING CHANGE:**


kdy1 (approved):
Thank you!

kdy1 (changes requested):
Can you rebase?

Issue #1063, opened 2023-08-13
[to do] implement type param indexing


sunrabbit123:
remove next and implement this
https://github.com/dudykr/stc/blob/main/crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs#L1319-L1329

```ts
// @strict: true


function l<T extends {}, P extends keyof T>(s: string, tp: T[P]): void {
    tp = s;
}
function m<T extends { a: number }, P extends keyof T>(s: string, tp: T[P]): void {
    tp = s;
}
```

T[P] should be number, but etc emit any;

PR #1067, opened 2023-08-13
feat: Implement property access about non-primitive constraints of IAT

sunrabbit123:
**Description:**
```ts
function l<T extends {}, P extends keyof T>(s: string, tp: T[P]): void {
    tp = s;
}
function m<T extends { a: number }, P extends keyof T>(s: string, tp: T[P]): void {
    tp = s;
}
function f<T extends object, P extends keyof T>(s: string, tp: T[P]): void {
    tp = s;
}
```

**BREAKING CHANGE:**

#1063 

---

## status

```ts
function f12<T, U extends T>(x: T, y: Partial<U>, k: keyof T) {
    x[k] = y[k];  // Error
    y[k] = x[k];  // Error
}
```

- [x] type inference
- [x] merge #1069 
- [x] merge #1070 
- [x] merge #1071 
- [x] merge #1072 
- [x] implement assign type(`T[keyof T] = U[keyof T] | undefined` , `U[keyof T] | undefined = T[keyof T]`)
- [x] merge #1073 
- [ ] remove increased extra error...


sunrabbit123:
I can't help but feel like I'm approaching this the wrong way.

TypeParam can't be keyofed beforehand because when it receives an argument value, it receives a value that inherits the constraints.

But in some cases it does

Why?

crates/stc_ts_type_checker/tests/conformance/types/keyof/keyofAndIndexedAccess.error-diff.json
```diff
@@ -11,6 +11,7 @@
     "TS2345": [
       96,
       132,
+      162,
```
sunrabbit123, line 14: when call JsFunction, we overwrite generic

```ts

function f33<S extends Shape, K extends keyof S>(shape: S, key: K) {
    let name = getProperty(shape, "name");
    let prop = getProperty(shape, key);
    return prop;
}

function f34(ts: TaggedShape) {
    let tag1 = f33(ts, "tag");
    let tag2 = getProperty(ts, "tag");
}
```

Change S to TaggedSchema, K to "tag"

crates/stc_ts_type_checker/tests/conformance/types/mapped/mappedTypeRelationships.error-diff.json
```diff
@@ -30,17 +16,32 @@
...
   },
   "extra_errors": {
-    "TS2322": 5
+    "TS2322": 8
   },
   "extra_error_lines": {
     "TS2322": [
+      18,
+      24,
+      28,
```
sunrabbit123, line 40: emit 2536 error

```ts
// @strictNullChecks: true
// @declaration: true

function f6<T, U extends T, K extends keyof U>(x: T, y: U, k: K) {
    x[k] = y[k];  // Error
    y[k] = x[k];  // Error
}
```

kdy1 (approved):

PR #1069, opened 2023-08-17
fix: a lot of test rule is not work

sunrabbit123:
**Description:**

When comparing TSC to error, we took the error results from ACTUAL.

In actual, the rules for each file were not applied properly,
As a result, tests like strictNullCheck didn't work correctly.

To fix this, we unified the env import code within the tests.

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

PR #1070, opened 2023-08-19
fix: Fix panic from `parse().unwrap()` at generic inference

sunrabbit123:
**Description:**

```rs
 if r.is_num() {
      match src.parse() {
          Ok(v) => {
              return Type::Lit(LitType {
                  span,
                  lit: RTsLit::Number(RNumber { span, value: v, raw: None }),
                  metadata: Default::default(),
                  tracker: Default::default(),
              })
          }
          Err(..) => {
              return Type::Keyword(KeywordType {
                  span,
                  kind: TsKeywordTypeKind::TsNumberKeyword,
                  metadata: Default::default(),
                  tracker: Default::default(),
              })
          }
      }
  }
```

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

PR #1071, opened 2023-08-19
Fix: add assign feature about indexed access type

sunrabbit123:
**Description:**

```ts
type A = 1 | 2;
function f<T extends A>(a: T): A & T { return a; } // Shouldn't error
```

**BREAKING CHANGE:**


kdy1 (approved):
Thsnks!

PR #1072, opened 2023-08-19
feat: Support accessing nom primitive constraint of an empty object

sunrabbit123:
**Description:**

```ts
function l<T extends {}, P extends keyof T>(s: string, tp: T[P]): void {
    tp = s;
}
```

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

PR #1073, opened 2023-08-20
feat: Union the type of an optional property of a mapped type with `undefined`

sunrabbit123:
**Description:**

```ts
function f10<T>(x: T, y: Partial<T>, k: keyof T) {
    x[k] = y[k];  // Error
    y[k] = x[k];
}
```

**BREAKING CHANGE:**


kdy1 (approved):
Thanks!

Issue #1081, opened 2023-09-08
Timer logging should be switched to a flag.


sunrabbit123:
1. the function is platform dependent
2. if you want to bind to multiple platforms, this function can be a headache.
3. when not in debug mode, it times but does not log.

https://github.com/dudykr/stc/blob/main/crates/stc_ts_type_checker/src/lib.rs#L91-L107

sunrabbit123:
part of resolve #1085 

not implemented flag system

moaoa:
hi there I am new to rust 
```
  pub fn check(&self, entry: Arc<FileName>) -> ModuleId {
        let timer = PerfTimer::tracing_debug();

        let modules = self.module_loader.load_module(&entry, true).expect("failed to load entry");
        #[cfg(debug_assertions)]
        timer.log(&format!("Loading of `{}` and dependencies", entry));
        let timer = PerfTimer::tracing_debug();

        self.analyze_module(None, entry.clone());
        #[cfg(debug_assertions)]
        timer.log(&format!("Analysis of `{}` and dependencies", entry));

        modules.entry.id
    }
```
should the changes be like this?
@sunrabbit123

moaoa:
or should i use #[cfg(feature = "debug")] ?

sunrabbit123:
It's a similar feeling, but I hope that the ratings for debug, information, error, etc. will be divided

moaoa:
Do you mean we need to create another flag for logging?

moaoa:
Do you mean we need to create another flag for logging?

Something like this:

#[cfg(feature = "log")]

sunrabbit123:
I know that you can't assign a value to a feature flag.
Is there a workaround?

What I'm envisioning is to divide the log levels and ignore logs below a certain level by a certain value at compile time.


Of course, the above is just my wishful thinking.

The purpose of the feature is to make it work without bugs in NAPI, WASM, and not to have logs that are not needed.


And just in case it's confusing, we'll mention the
When I refer to flags in the text, I mean argument values for something, not rust feature flags.
Sure, it could be a rust feature flag, but I think it has limited functionality.

kdy1:
Yeah, feature flags are quite limited. We may create our own macro for logging which invokes macros from tracing.
If we do so, we will have much more detailed control, even with  cargo feature flags.


cargo feature:

 - `stc_logger/log-type-creation`
 - `stc_logger/info` (maybe)

moaoa:
[...]

Issue #1021, opened 2023-05-12
Would you be willing to provide an STC playground?


sunrabbit123:
When creating a contribution or issue, it would be nice to have a place to look up the most recent STC and write test cases more easily.
You can more easily check if an existing case is a working case or not.

Like how we use Typescript Playground to check types.

kdy1:
I want to, but this is a very complex task..

DavidHancu:
@kdy1 I'd be down to work on this if we could get hosting for it.

phoenix-ru:
What needs to be done to make it possible? Afaik, there is an LSP setup with VSCode. Maybe this can be done the same way as SWC playground with the nightly build of STC as a plugin?

kdy1:
Code would be almost similar, but we should publish Wasm files

sunrabbit123:
Do you need to do something like #300 ?
If so, do you use wasm-bindgen instead of napi-rs?

PR #1082, opened 2023-09-08
feat: (PoC) impl wasm support for playground.

AcrylicShrimp:
**Description:**

This PR implements checker behaviors for `wasm` binding. By supporting wasm, implementing playground will become very easy. I also implemented PoC of playground and it works well. I'll contribute on that when this PR is ready and merged.

Due to the constraints of wasm, there is a few APIs that should not be used. Below is the list.

- `std::time::Instant`
- `mimalloc_rust` from `swc_node_base`
- All file-system related APIs.

We dropped some code to avoid use of the APIs, and this PR contains those changes. I marked this PR as draft; we have to manage them somehow before merge it.

**Related issue:**
#300
#1021 

**Task**

- [ ] #1081


sunrabbit123:
@kdy1 If this issue passes, I think we should create an stc-playground repository.

crates/stc_utils/src/lib.rs
```diff
@@ -1,8 +1,5 @@
 #![feature(never_type)]
 
-/// Use good memory allocator.
-extern crate swc_node_base;
```
sunrabbit123, line 4: @kdy1 

I discovery this code is unused
So I think it's safe to delete it, what do you think?

> `swc` uses this when attempting to bind to the napi.rs.

kdy1: IIRC, `swc_node_base` is about using a faster mallocator

crates/binding_wasm/src/lib.rs
```diff
@@ -1,14 +1,146 @@
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CheckAndAnnotateOutput {
+    error: String,
+}
+
+pub struct PassThroughWriter {
+    pub buf: Arc<RefCell<BufWriter<Vec<u8>>>>,
```
kdy1, line 88: This should be `Arc<Mutex<Vec<u8>>>`.

 - `BufWriter<Vec<u8>>` has nothing good over `Vec<u8>` becuase it's already in memory.
 - `Arc<RefCell<T>>` is unsound, and it's `unsafe.`

crates/binding_wasm/src/lib.rs
```diff
@@ -1,14 +1,146 @@
...
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        self.buf.borrow_mut().write(buf)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        self.buf.borrow_mut().flush()
+    }
+}
+
+unsafe impl Send for PassThroughWriter {}
```
kdy1, line 101: This can be removed

kdy1 (changes requested):

sunrabbit123:
@AcrylicShrimp
Can you resolve clippy error?

AcrylicShrimp:
I resolved all problems. Should I do rebase? @sunrabbit123

sunrabbit123:
Resolved with [commit(7dc35c8a8a0d64703ab152e)](https://github.com/dudykr/stc/pull/1082/commits/c32f50de9806222187dc35c8a8a0d64703ab152e). Thank you.

May I ask for a review? @kdy1

sunrabbit123:
I checked the related issue
I think a name similar to `stc_playground_wasm` would be better than `binding-wasm`.

kdy1:
>  I think a name similar to stc_playground_wasm would be better than binding-wasm.


I agree with this. In my experience, we will want many APIs for the playground not exposed to normal wasm bindings

kdy1 (approved):
Thanks!

PR #1084, opened 2023-09-09
feat: Improve optional chaining validation

sunrabbit123:
**Description:**

**BREAKING CHANGE:**


crates/stc_ts_type_checker/tests/conformance/controlFlow/controlFlowOptionalChain.error-diff.json
```diff
@@ -25,6 +25,8 @@
       223,
       241,
       244,
+      370,
```
sunrabbit123, line 28: else, there should be separate logic in there, but putting it in false fact doesn't seem to work.
What am I missing?

sunrabbit123:
```ts
type Thing = { foo: string | number, bar(): number, baz: object };

function f16(o: Thing | undefined) {
    if (o?.foo === undefined) {
        o.foo;  // Error
    }
    else {
        o.foo;
    }
    if (o?.foo !== undefined) {
        o.foo;
    }
    else {
        o.foo;  // L370
    }
    if (o?.foo == undefined) {
        o.foo;  // Error
    }
    else {
        o.foo;
    }
    if (o?.foo != undefined) {
        o.foo;
    }
    else {
        o.foo;  // L382
    }
}
```

kdy1 (approved):
Thank you! Can you rebase?

sunrabbit123:
sir, i rebased it

kdy1 (approved):

PR #1086, opened 2023-09-11
feat: Remove panic when `main_source` contains reference comment

sunrabbit123:
**Description:**

```rs
let main_src = Arc::new(fs::read_to_string(file_name).unwrap());
// Postpone multi-file tests.
if main_src.contains("<reference path") {
    panic!("`<reference path` is not supported yet");
}
```
remove it for next

```tsx
/// <reference lib="dom" />
```

**Deep Dive**
this code wrote #487

I'm guessing it was used as a pass-through because it was a test case that was too much for the code to handle at the time.
And I don't think it matters if you delete that code now.


sunrabbit123:
If you think it should be implemented from `reference`,
I think it's better to convert that pr to draft or create an issue after merging that pr.

kdy1 (approved):

PR #1087, opened 2023-09-13
fix: Fix parsing of numeric literals like `0x9E`

sunrabbit123:
**Description:**

```ts
type TBigInt2 = "0x10" extends `${infer N extends bigint}` ? N : never; // bigint (not round-trippable)
type TBigInt3 = "0o10" extends `${infer N extends bigint}` ? N : never; // bigint (not round-trippable)
type TBigInt4 = "0b10" extends `${infer N extends bigint}` ? N : never; // bigint (not round-trippable)
```


kdy1 (dismissed):
Can you rebase?

kdy1 (approved):

PR #1089, opened 2023-09-17
feat: resolve panic cause by importMeta type

sunrabbit123:
**Description:**

```ts
declare global { interface ImportMeta {foo?: () => void} };

if (import.meta.foo) {
  import.meta.foo();
}
```

**BREAKING CHANGE:**


crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -333,15 +333,6 @@ impl Analyzer<'_, '_> {
         ty.freeze();
         self.cur_facts.assert_clone_cheap();
 
-        if !self.config.is_builtin && !ty.is_any() {
```
kdy1, line 336: This should not be removed.

kdy1: Please change the span of the generated `Type` instead

kdy1 (changes requested):

crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
sunrabbit123: span of builtin type eqaul dummy_span
chnage to span of `import.meta` keyword?
@kdy1

kdy1: Yes

kdy1 (approved):

PR #1095, opened 2023-10-04
chore: Update rust-toolchain to nightly-2023-10-04

simonbuchan:
**Description:**

Just updating the toolchain to the latest to get my toes wet. Everything I could try still seems to act the same as far as I can tell, just needed to bump `proc-macro2` for a removed feature, and I set `workspace.resolver` explicitly due to what seems like a new warning - since there aren't any top-level dependencies it seems like this didn't change any dependencies?

**BREAKING CHANGE:**

Shouldn't be anything other than possibly affecting MSRV if there is any policy there.

Last update was #1034


simonbuchan:
Ah, of course I should have expected `fmt` / `clippy` changes :| They seem harmless enough?

simonbuchan:
Let me know if you don't like any of the fixes!

crates/stc_ts_builtin_macro/src/lib.rs
```diff
@@ -5,20 +5,20 @@
...
+use swc_common::{comments::SingleThreadedComments, sync::Lrc, FilePathMapping, SourceMap};
 use swc_ecma_parser::{lexer::Lexer, Parser, StringInput, Syntax, TsConfig};
 use swc_macros_common::{call_site, print};
 use syn::{punctuated::Punctuated, Token};
 
 #[proc_macro]
 pub fn builtin(_: proc_macro::TokenStream) -> proc_macro::TokenStream {
     swc_common::GLOBALS.set(&swc_common::Globals::new(), || {
-        let cm = Arc::new(SourceMap::new(FilePathMapping::empty()));
+        let cm = Lrc::new(SourceMap::new(FilePathMapping::empty()));
```
simonbuchan, line 21: It was complaining about `Arc` on a non-Send/Sync type, interestingly. I think that's probably a clippy bug? Strictly, you don't even need an rc at all, but the docs for SourceMap suggest using this alias, which suppresses the lint.

crates/stc_ts_file_analyzer/src/analyzer/assign/tpl.rs
```diff
@@ -27,6 +27,7 @@ impl Analyzer<'_, '_> {
     /// orders.
     ///
     /// After splitting, we can check if each element is assignable.
+    #[allow(clippy::needless_pass_by_ref_mut)]
```
simonbuchan, line 30: Complaining about `data` not being mutably used here (though it's not even actually used, there's an `allow(unused)`)

I didn't want to change APIs here, but that's probably more correct.

crates/stc_ts_file_analyzer/src/analyzer/function/mod.rs
```diff
@@ -348,11 +348,7 @@ impl Analyzer<'_, '_> {
             //         true,
             //         // Allow overriding
             //         true,
-            //     ) {
-            //         Ok(()) => {}
-            //         Err(err) => {
-            //             self.storage.report(err);
-            //         }
+            //     ) { Ok(()) => {} Err(err) => { self.storage.report(err); }
```
simonbuchan, line 351: Seems uglier to me? Maybe there's a rustfmt option.

crates/stc_ts_file_analyzer/src/analyzer/generic/mod.rs
```diff
@@ -111,10 +111,7 @@ impl Analyzer<'_, '_> {
         #[cfg(debug_assertions)]
         let _tracing = dev_span!("infer_arg_types");
 
-        warn!(
-            "infer_arg_types: {:?}",
-            type_params.iter().map(|p| format!("{}, ", p.name)).collect::<String>()
-        );
+        warn!("infer_arg_types: {:?}", type_params.iter().map(|p| &p.name).join(", "));
```
simonbuchan, line 114: Technically a different format (no trailing `, `), but it was complaining about the inefficient format, and this is much clearer.

kdy1 (approved):
Thanks! All looks good

PR #1097, opened 2023-10-05
Draft: chore: replace scripts with xtask pattern

simonbuchan:
**Description:**

Migrates scripts to use xtask pattern.

See discussion in #1096


simonbuchan:
Not dead yet! I got distracted trying to figure out why `mimalloc-rust` (used by `swc_node_base` on non-linux platforms) was being all crashy, but that should be a separate PR if I do figure it out.

This is pretty close for what's implemented: the enabled tests for `stc_ts_{file_analyzer,type_checker}`.

The main thing left I know about is that `cargo xtask auto-unignore` is generating a bunch of `.*.tsc-errors.json` files outside of `tests/tsc`, so they aren't getting cleaned up. I have no idea what I'm doing differently to the shell script here, unfortunately.

kdy1:
I can take a look when I have time

simonbuchan:
Thanks, I'll keep at it anyway. I'm pretty sure it's effectively the same command line (modulo order), so it's most likely something to do with current directory, variables, me using volta or the like. I just can't track down why that would result in what I'm seeing.

simonbuchan:
Ugh. Sorry for disappearing for like (checks) THREE WEEKS. Geez. I did poke at this a few times, but didn't figure out what could have been causing the last issue where `cargo xtask auto-unignore` was generating a bunch of `.stc-stderr` files.

Turned out to just be that `auto-unignore.sh` didn't set `UPDATE=1`. I have no idea why that makes a difference, there doesn't seem to be any reference to that variable in Rust source at least? Rather annoying that it was that simple.

And as I've said before, I'm happy to make any changes or additions you would prefer.

kdy1 (changes requested):
I think we can remove the replaced scripts

xtask/src/type_checker.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+    });
+
+    let status = child.wait()?;
+    ensure!(status.success(), "test conformance failed: {status}");
+
+    tests.sort();
+
+    use std::io::Write;
+    let mut file = std::fs::File::create("crates/stc_ts_type_checker/tests/conformance.pass.txt")?;
```
kdy1, line 78: Should be use a buffer here?

simonbuchan:
Hmm, cargo xtask test-checker did also panic on CI in "PR Maintenance / All passing tests should be enabled (pull_request) " ... might be something going on there? Not much detail.

simonbuchan:
Ah, it's expected to fail for now, but *also* it fails with mimalloc heap corruption on windows.

I fixed it not updating conformance.pass.txt if it fails, and also tossed in a dot progress reporter and stopped it filtering out `.tsx` test passes. Let me know if either of those seem bad to you!

kdy1:
It's fine, but CI failed.

simonbuchan:
Yeah. Tested on WSL Ubuntu and it ran through there. Any idea how to get
the panic message at least?

PR #1099, opened 2023-10-10
fix: Import local type from module and namespace

sunrabbit123:
**Description:**
```ts

declare module "foo2" {
    namespace Bar {
        interface I {
            a: string;
            b: number;
        }
    }

    namespace Baz {
        interface J {
            a: number;
            b: string;
        }
    }

    class Bar {
        item: Bar.I;
        constructor(input: Baz.J);
    }
}

let y: import("foo2").Bar.I = { a: "", b: 0 };
```

**BREAKING CHANGE:**


crates/stc_ts_file_analyzer/src/analyzer/expr/mod.rs
```diff
@@ -3053,16 +3053,22 @@ impl Analyzer<'_, '_> {
                             if let Some(vars) = exports.vars.get(sym).cloned() {
                                 return Ok(vars);
                             }
+
+                            for (id, tys) in exports.private_types.iter() {
+                                if id.sym().eq(sym) {
```
kdy1, line 3058: Please use `==` instead

sunrabbit123: Is there a particular reason other than readability?

sunrabbit123, line 3058:
```suggestion
                                if *id.sym() == *sym {
```

crates/stc_ts_file_analyzer/src/analyzer/scope/mod.rs
```diff
@@ -2247,6 +2252,16 @@ impl Expander<'_, '_, '_> {
 
                             Type::Mapped(m) => {}
 
+                            Type::Namespace(..) => {
+                                dbg!(&t);
```
kdy1, line 2256: Please remove this

sunrabbit123: !! my mistake
I'll fix it right away
thank you

kdy1 (approved):
Thanks!

PR #1103, opened 2023-11-14
feat: Remove extra error at `interfaceExtendsObjectIntersectionErrors`

sunrabbit123:
**Description:**

```ts
interface I31<T> extends T { x: string }
```

I have a concern.

The following functions cause duplicate error handling due to duplicate function calls.

```rs
#[validator]
impl Analyzer<'_, '_> {
    fn validate(&mut self, d: &RTsInterfaceDecl) -> VResult<Type> {
        let ty = self.with_child(ScopeKind::Flow, Default::default(), |child: &mut Analyzer| -> VResult<_> {
            match &*d.id.sym {
                "any" | "void" | "never" | "unknown" | "string" | "number" | "bigint" | "boolean" | "null" | "undefined" | "symbol" => {
                    child.storage.report(ErrorKind::InvalidInterfaceName { span: d.id.span }.into());
                }
                _ => {}
            }

            let mut ty = Interface {
                span: d.span,
                name: d.id.clone().into(),
                type_params: try_opt!(d.type_params.validate_with(&mut *child).map(|v| v.map(Box::new))),
                extends: d.extends.validate_with(child)?.freezed(),
                body: d.body.validate_with(child)?,
                metadata: Default::default(),
                tracker: Default::default(),
            };
            child.prevent_expansion(&mut ty.body);
            ty.body.freeze();

            child.resolve_parent_interfaces(&d.extends, true);
            child.report_error_for_conflicting_parents(d.id.span, &ty.extends);
            child.report_error_for_wrong_interface_inheritance(d.id.span, &ty.body, &ty.extends);

            let ty = Type::Interface(ty).freezed();

            Ok(ty)
        })?;

        // TODO(kdy1): Recover
        self.register_type(d.id.clone().into(), ty.clone());

        Ok(ty)
    }
}
```

`child.resolve_parent_interfaces` and `child.report_error_for_wrong_interface_inheritance` call `report_error_for_unresolved_type`
this fn cause `ErrorKind::TypeNotFound`

This PR only clogs the hole.

If there seems to be a need for fundamental improvement, please open up the issue

**BREAKING CHANGE:**

### DIFF
--- a/crates/stc_ts_file_analyzer/src/analyzer/convert/interface.rs
+++ b/crates/stc_ts_file_analyzer/src/analyzer/convert/interface.rs
@@ -62,6 +62,11 @@ impl Analyzer<'_, '_> {
             };
 
             if let Err(err) = res {
+                if err.code() == 2304 {
+                    // TypeNotFound
+                    return;
+                }
+
                 self.storage.report(
                     ErrorKind::InvalidInterfaceInheritance {
                         span,
--- a/crates/stc_ts_file_analyzer/tests/tsc/types/interface/interfaceDeclaration/1.ts
+++ b/crates/stc_ts_file_analyzer/tests/tsc/types/interface/interfaceDeclaration/1.ts
@@ -1,0 +1,3 @@
+// @strictNullChecks: true
+
+interface I31<T> extends T { x: string }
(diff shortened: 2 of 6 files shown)


kdy1 (approved):

kdy1:
I think `report_error_for_wrong_interface_inheritance` should not call `report_error_for_unresolved_type`

sunrabbit123:
umm... the flow of the function is as follows.

`report_error_for_wrong_interface_inheritance` -> `self.type_of_ts_entity_name` -> `type_of_ts_entity_name_inner` -> ` report_error_for_unresolved_type`

kdy1:
No, but we may use `ctx` to pass option to ignore resolving failure

sunrabbit123:
I'll put it up as an issue for refactoring.

I can't think of a good way now.

Because a simple if statement returns Ref, which breaks the logic, is worse than not in consequence.

thx.

